PR #4, opened 2022-09-15
fix: createAccount response fields and updated test cases and the ope…

jonesho:
…nrpc spec

### DIFF
--- a/packages/starknet-snap/openrpc/starknet_snap_api_openrpc.json
+++ b/packages/starknet-snap/openrpc/starknet_snap_api_openrpc.json
@@ -1036,9 +1036,6 @@
           "address": {
             "$ref": "#/components/schemas/ADDRESS",
             "description": "The address of the deploying account"
-          },
-          "code": {
-            "description": "Response code of the deploy txn"
           }
         }
       },
--- a/packages/starknet-snap/package.json
+++ b/packages/starknet-snap/package.json
@@ -51,7 +51,7 @@
     "ethereum-unit-converter": "^0.0.17",
     "sinon": "^13.0.2",
     "sinon-chai": "^3.7.0",
-    "starknet": "^4.5.0"
+    "starknet": "^4.6.0"
   },
   "publishConfig": {
     "access": "public",
--- a/packages/starknet-snap/snap.manifest.json
+++ b/packages/starknet-snap/snap.manifest.json
@@ -7,7 +7,7 @@
     "url": "https://github.com/ConsenSys/starknet-snap.git"
   },
   "source": {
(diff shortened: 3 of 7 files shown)


Julink-eth (commented):
I'm getting the error : "Problem found: Error: -32602: More than 64 digits found: 76 at line 1 column 149" after calling createAccount.

wantedsystem:
Hello @jonesho  i'm getting this error when switching to mainnet after deploying a contract 
`MetaMask - RPC Error: The given token address and user address need to be non-empty string, got: {"tokenAddress":"0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7","chainId":"0x534e5f4d41494e"} `

Screenshot 2022-09-15 at 14 49 32

jonesho:
@wantedsystem that was the exact issue I want to fix by this PR, have u tried checkout the latest changes of the feature branch since I didn't delete it from my previous PR and re-used it again for this PR, also could you pls also makand e sure the Dapp has disconnected and reconnected to the snap? I couldn't reproduce this anymore from my side which I did use new MM seed phrase and was able to deploy new accounts in both mainnet and testnet

Julink-eth:
> @Julink-eth I believe it's more likely the Infura RPC temporary issue at that moment, I can't reproduce it from my side, I used a new MM seed phrase and was able to deploy new acc in both testnet and mainnet

I tried again with a fresh flask and I still have the same issue

jonesho:
@Julink-eth I think you're hitting this issue: https://github.com/0xs34n/starknet.js/issues/285
This issue should be fixed by starknet.js v4.5.0 which is the current version used in our snap, just to check did you checked out the latest of this feature branch (SF-384-upgrade-starknet-js-and-make-infura-rpc-endpoints-as-default), build and run it locally with yarn install, then disconnect and reconnect the Dapp to the locally-running snap?

This is the command I used for local build and run the snap:
`yarn workspace @consensys/starknet-snap install && yarn workspace @consensys/starknet-snap build:clean && yarn workspace @consensys/starknet-snap serve`

jonesho:
@Julink-eth It seems like the snap you run is using the old version of starknet.js

Julink-eth:
> @Julink-eth It seems like the snap you run is using the old version of starknet.js

After pulling the latest changes, it's working indeed, thanks Jones.

Julink-eth (approved):

wantedsystem:
Hello @jonesho it's working now but i got another error in the console
Screenshot 2022-09-16 at 09 43 11
i don't know if it's linked but it's working fine

jonesho:
@wantedsystem it's not related to the implementation of our snap, could be from the Infura RPC side, I also saw this from time to time without impacting the functionality

wantedsystem (approved):

PR #10, opened 2022-09-21
feat: ensure txns in snap state have the same status and timestamp as…

jonesho:
… that of Voyager


jonesho:
The improvement we made here is to always treat the txns from the Voyager as the source of truth and update the same txns found in the snap state with the same status and timestamp as that of Voyager.

Julink-eth (approved):

wantedsystem (approved):

PR #33, opened 2022-10-25
chore: release prod

wantedsystem:


onetrickwolf:
@wantedsystem can merge this https://github.com/ConsenSys/starknet-snap/pull/32 and I'll do a release with the new process since staging is behind here and the new process fixes this.

PR #39, opened 2022-11-21
feat: temporarily reverted back to use StarkNet feeder gateway APIs a…

jonesho:
…s default provider


jonesho:
Temporarily reverted back to use StarkNet feeder gateway as default provider, until the starknet.js is upgraded with version compatible with Infura pathfinder v0.4.0. Also add support of StarkNet Goerli 2 testnet in snap JSON-RPC APIs.

Julink-eth (approved):

wantedsystem (approved):

onetrickwolf (approved):

PR #50, opened 2023-01-30
feat: added support for MM flask v10.24.1 and added starkNet_estimate…

jonesho:
…AccountDeployFee

### DIFF
--- a/packages/wallet-ui/src/components/ui/organism/Header/SendSummaryModal/SendSummaryModal.view.tsx
+++ b/packages/wallet-ui/src/components/ui/organism/Header/SendSummaryModal/SendSummaryModal.view.tsx
@@ -232,7 +232,7 @@ export const SendSummaryModalView = ({ address, amount, chainId, closeModal }: P
           </RightSummary>
         </Summary>
         {!estimatingGas && <TotalAmount>Maximum fees: {gasFeesAmount} ETH</TotalAmount>}
-        {gasFees.includeDeploy && <IncludeDeploy>*Fees include a one-time deployment fee</IncludeDeploy>}
+        {gasFees.includeDeploy && <IncludeDeploy>*Fees include an one-time deployment fee</IncludeDeploy>}
         <Summary>
           <LeftSummary>
             <PopperTooltip placement="right" closeTrigger="hover" content="Amount + Fee">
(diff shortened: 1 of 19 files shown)


wantedsystem (dismissed):

Julink-eth (dismissed):

packages/wallet-ui/src/components/ui/organism/Header/SendSummaryModal/SendSummaryModal.view.tsx
```diff
@@ -232,7 +232,7 @@ export const SendSummaryModalView = ({ address, amount, chainId, closeModal }: P
           </RightSummary>
         </Summary>
         {!estimatingGas && <TotalAmount>Maximum fees: {gasFeesAmount} ETH</TotalAmount>}
-        {gasFees.includeDeploy && <IncludeDeploy>*Fees include a one-time deployment fee</IncludeDeploy>}
+        {gasFees.includeDeploy && <IncludeDeploy>*Fees include an one-time deployment fee</IncludeDeploy>}
```
Julink-eth, line 235: I think the correct spelling is actually "a" here 
https://www.quora.com/Which-is-correct-a-one-or-an-one#:~:text=It's%20'a%20one%2Dtime%20opportunity,sound%2C%20not%20an%20o%20sound.

Julink-eth:
I have an issue if I update my flask and I already had the flask + snap installed.
The dapp asks me to install the latest snap which is 1.2.1 here but then I get the error : "Failed to decrypt snap state, the state must be corrupted." once installed
image

If I uninstall the snap and reinstall, it works fine.

packages/wallet-ui/src/components/ui/organism/Header/SendSummaryModal/SendSummaryModal.view.tsx
jonesho: @Julink-eth oh u r right, nice catch!

jonesho:
@Julink-eth pushed another commit, for the install snap error, I guess its because your existing and new snap version are both in v1.2.1 and the dApp was calling the old snap which is not compatible with Flask v10.25, and hence u saw the "snap state" error message.

Julink-eth:
> I have an issue if I update my flask and I already had the flask + snap installed. The dapp asks me to install the latest snap which is 1.2.1 here but then I get the error : "Failed to decrypt snap state, the state must be corrupted." once installed image
> 
> If I uninstall the snap and reinstall, it works fine.

Nevermind https://www.notion.so/0-27-1-24767c05a7684851ab856fbe8a2370b5

Julink-eth:
> @Julink-eth pushed another commit, for the install snap error, I guess its because your existing and new snap version are both in v1.2.1 and the dApp was calling the old snap which is not compatible with Flask v10.25, and hence u saw the "snap state" error message.

Yes exactly and I think the only thing we could do is display a modal for the user to reinstall their snap because they are going to be stuck otherwise

PR #59, opened 2023-02-10
feat: improved decimals display + fix max function in the send modal

Julink-eth:
After discussing with Arthur, I changed how we display the decimals in the dapp so that it is more user friendly and "smart".
I also fixed an issue we had with the max button that would sometimes trigger the amount exceeds user balance error.


wantedsystem (approved):
LGTM

PR #92, opened 2023-07-07
fix(SF-490): prevent markdown injection

Akaryatrh:
Fixes https://consensyssoftware.atlassian.net/browse/SF-490
- ~~prevent markdown injection on contract call data~~
- use copyable component to display contract call data
- throw properly when contract call data format can not be converted to a bigNumber


packages/starknet-snap/src/sendTransaction.ts
```diff
@@ -58,15 +57,14 @@ export async function sendTransaction(params: ApiParams) {
...
-
     const response = await wallet.request({
       method: 'snap_dialog',
       params: {
         type: DialogType.Confirmation,
         content: panel([
           heading('Do you want to sign this transaction ?'),
           text(`It will be signed with address: ${senderAddress}`),
-          text(signingTxnText),
+          copyable(signingTxnText),
```
stanleyyconsensys, line 67: straightly using copyable will direct output the line break text "\n\n" rather than doing a line break

we may need to divide the text into different component, e.g:
text("**heading:**"),
copyable(val),...

i have created a PR to address the layout, you can check it out and see if it is okay
https://github.com/Consensys/starknet-snap/pull/105/files

stanleyyconsensys (changes requested):

PR #99, opened 2023-07-11
feat(SF-491): add length validator for adresses

Akaryatrh:


packages/starknet-snap/test/utils/starknetUtils.test.ts
```diff
@@ -45,3 +46,18 @@ describe('Test function: callContract', function () {
...
 });
+
+describe('Test function: validateAndParseAddress', function () {
+
+  it.skip('should call initial validateAndParseAddress when addresses have proper length', async function () {
+    const validateAndParseAddressSpy = sinon.spy(starknet, 'validateAndParseAddress');
+    utils.validateAndParseAddress(account1.address);
+    utils.validateAndParseAddress(account1.addressSalt);
+    expect(validateAndParseAddressSpy).to.have.been.called;
+  });
```
Akaryatrh, line 57: I need help on this one: Starknet package functions are exported in a way that they can not be spied (because `configurable` property is not enabled by `Typescript`). So, this test is failing and returns an error (`TypeError: Attempted to wrap undefined property validateAndParseAddress as function`).

if you have any alternative, that would be helpful 🙂

stanleyyconsensys: the easier way is 
- create a const method and export in starknetUtils.ts
import { validateAndParseAddress as validateAndParseAddress_org } from 'starknet';
e.g export const _validateAndParseAddress = validateAndParseAddress_org;
- in your method, call _validateAndParseAddress instead (if you have to name it as same as starknet
- in your test, mock (utils, '_validateAndParseAddress')

Akaryatrh: Super simple solution and works nicely, thanks a lot!

PR #101, opened 2023-07-11
fix(SF-499): rearrange deps and dev deps

Akaryatrh:


packages/wallet-ui/package.json
```diff
@@ -27,7 +26,6 @@
     "react-qr-code": "^2.0.7",
     "react-redux": "^8.0.1",
     "redux-persist": "^6.0.0",
-    "semver": "^7.3.7",
```
stanleyyconsensys, line 30: semver is using in useStarkNetSnap.ts -> oldVersionDetected() , should not removed

stanleyyconsensys (changes requested):

PR #104, opened 2023-07-13
fix(SF-498): add logger wrap console log

stanleyyconsensys:
- add logger wrap console log
- change console.{method} to logger.{method}
- add unit test for logger
- update production cicd


packages/wallet-ui/src/services/useStarkNetSnap.ts
```diff
@@ -31,6 +31,11 @@ export const useStarkNetSnap = () => {
   const snapId = process.env.REACT_APP_SNAP_ID ? process.env.REACT_APP_SNAP_ID : 'local:http://localhost:8081/';
   const snapVersion = process.env.REACT_APP_SNAP_VERSION ? process.env.REACT_APP_SNAP_VERSION : '*';
   const minSnapVersion = process.env.REACT_APP_MIN_SNAP_VERSION ? process.env.REACT_APP_MIN_SNAP_VERSION : '1.7.0';
+  const debugLevel = process.env.REACT_APP_DEBUG_LEVEL ? process.env.REACT_APP_DEBUG_LEVEL : 'all';
```
Akaryatrh, line 34: [optional] or simpler: 
```suggestion
  const debugLevel = process.env.REACT_APP_DEBUG_LEVEL || 'all';
```

Akaryatrh (changes requested):

packages/wallet-ui/src/services/useStarkNetSnap.ts
stanleyyconsensys: i have to change it to` process.env.REACT_APP_DEBUG_LEVEL !== undefined ? process.env.REACT_APP_DEBUG_LEVEL : 'all';`

REACT_APP_DEBUG_LEVEL can be '0' , which also means off

PR #121, opened 2023-08-23
feat: migrate metamask flask handle to metamask

stanleyyconsensys:
- add metamask detect and alert model
- update flask content to metamask content
- update fox icon
- add starknet icon to snap
- fix lint style
- handle multi wallets

ticket: https://consensyssoftware.atlassian.net/jira/software/projects/SF/boards/472?selectedIssue=SF-517


packages/wallet-ui/src/components/ui/organism/NoFlaskModal/NoFlaskModal.style.ts
```diff
@@ -1,6 +1,6 @@
 import styled from 'styled-components';
```
wantedsystem, line 1: We still want to use snaps with Flask ? if it's not the case you can add a comment to remember that we need to remove this component after releasing the snaps in production

wantedsystem (dismissed):
LGTM

packages/wallet-ui/src/components/ui/organism/NoFlaskModal/NoFlaskModal.style.ts
stanleyyconsensys: noted, will add a comment to remove this component after release

PR #158, opened 2023-10-31
feat: sign transaction ongoing

wantedsystem:


packages/starknet-snap/src/signTransaction.ts
```diff
@@ -0,0 +1,27 @@
+import { Signature, Signer, stark } from 'starknet';
+import { ApiParams, SignTransactionParams } from './types/snapApi';
+import { getKeysFromAddress } from './utils/starknetUtils';
+import { getNetworkFromChainId } from './utils/snapUtils';
+
+export async function signTransaction(params: ApiParams): Promise<Signature> {
+  const { state, keyDeriver, requestParams } = params;
+  const requestParamsObj = requestParams as SignTransactionParams;
+  const userAddress = requestParamsObj.userAddress;
```
stanleyyconsensys, line 9: address validation should be added

packages/starknet-snap/src/signTransaction.ts
```diff
@@ -0,0 +1,27 @@
...
+  const network = getNetworkFromChainId(state, requestParamsObj.chainId);
+  const { privateKey } = await getKeysFromAddress(keyDeriver, network, state, userAddress);
```
stanleyyconsensys, line 11: account should check if it has deployed

packages/starknet-snap/src/signTransaction.ts
```diff
@@ -0,0 +1,27 @@
...
+  const signer = new Signer(privateKey);
```
stanleyyconsensys, line 12: dialog with sign information should popup to ask user to consensus

packages/starknet-snap/src/signTransaction.ts
```diff
@@ -0,0 +1,27 @@
...
+  try {
+    const signatures = await signer.signTransaction(
+      requestParamsObj.transactions,
+      requestParamsObj.transactionsDetail,
+      requestParamsObj.abis,
+    );
+    const formattedSignatures = stark.signatureToDecimalArray(signatures);
+    const publicSigner = await signer.getPubKey();
+
+    return [publicSigner, ...formattedSignatures];
```
stanleyyconsensys, line 23: what is the use case to return a publicSigner?

packages/starknet-snap/src/signTransaction.ts
```diff
@@ -0,0 +1,27 @@
...
+  } catch (error) {
+    throw new Error('Error signing the transaction');
```
stanleyyconsensys, line 25: error should log and throw same err to client

packages/starknet-snap/src/types/snapApi.ts
```diff
@@ -146,3 +148,9 @@ export interface RpcV4GetTransactionReceiptResponse {
   execution_status?: string;
   finality_status?: string;
 }
+
+export interface SignTransactionParams extends BaseRequestParams, ExtractPrivateKeyRequestParams {
+  transactions: Call[];
```
stanleyyconsensys, line 153: the data is pass from frontend, call[] type is the underlying type, please refer to SendTransactionRequestParams

stanleyyconsensys: apply to the below following param as well

stanleyyconsensys (changes requested):
missing unit test

stanleyyconsensys (changes requested):

packages/starknet-snap/src/signTransaction.ts
bajpai244: Do we need to? Because this seems to be just for signing a transaction.

If it is from a UX perspective, then it makes sense!

packages/starknet-snap/src/signTransaction.ts
bajpai244: +1!

packages/starknet-snap/src/signTransaction.ts
bajpai244: The use I can think of is for verification, but it looks like it should probably work without it as well.

packages/starknet-snap/src/signTransaction.ts
stanleyyconsensys: TBC

packages/starknet-snap/src/signTransaction.ts
stanleyyconsensys: TBC

packages/starknet-snap/src/signTransaction.ts
stanleyyconsensys: TBC

packages/starknet-snap/src/types/snapApi.ts
stanleyyconsensys: TBC

packages/starknet-snap/index.html
```diff
@@ -507,6 +521,16 @@ <h1>Hello, Snaps!</h1>
...
     }
 
+    // here we call the snap's "starkNet_signTransaction" method
+    async function signTransaction(e) {
+      e.preventDefault(); // to prevent default form behavior
+
+      const transaction = document.getElementById('transactions').value;
+      const transactionsDetail = document.getElementById('transactionsDetail').value;
+
+      await callSnap('starkNet_signTransaction', { transaction, transactionsDetail });
```
stanleyyconsensys, line 531: the transaction and transactionsDetail may not able to passin to callSnap due to html form

packages/starknet-snap/src/signTransaction.ts
```diff
@@ -0,0 +1,25 @@
...
+import { getNetworkFromChainId } from './utils/snapUtils';
+
+export async function signTransaction(params: ApiParams): Promise<Signature> {
+  const { state, keyDeriver, requestParams } = params;
+  const requestParamsObj = requestParams as SignTransactionParams;
+  const userAddress = requestParamsObj.userAddress;
+  const network = getNetworkFromChainId(state, requestParamsObj.chainId);
+  const { privateKey } = await getKeysFromAddress(keyDeriver, network, state, userAddress);
+
+  try {
```
stanleyyconsensys, line 13: `try` should place it at the beginning of the method to catch all errors

packages/starknet-snap/src/utils/starknetUtils.ts
```diff
@@ -481,3 +486,22 @@ export const validateAndParseAddress = (address: num.BigNumberish, length = 63)
...
+
+export const signTransactions = async (
+  privateKey: string,
+  transactions: Call[],
+  transactionsDetail: InvocationsSignerDetails,
+  abis: Abi[],
+): Promise<Signature> => {
+  const signer = new Signer(privateKey);
+
+  try {
```
stanleyyconsensys, line 498: optional , try catch can be skip for utils function when it has catch at the snap method

packages/starknet-snap/src/signTransaction.ts
```diff
@@ -0,0 +1,25 @@
...
+    const signatures = await signTransactions(
+      privateKey,
+      requestParamsObj.transactions,
+      requestParamsObj.transactionsDetail,
+      requestParamsObj.abis,
+    );
+    return signatures;
+  } catch (error) {
+    console.error(`Problem found: ${error}`);
```
stanleyyconsensys, line 22: use logger.error instead

packages/starknet-snap/src/utils/starknetUtils.ts
```diff
@@ -481,3 +486,22 @@ export const validateAndParseAddress = (address: num.BigNumberish, length = 63)
...
+    const signatures = await signer.signTransaction(transactions, transactionsDetail, abis);
+    const formattedSignatures = stark.signatureToDecimalArray(signatures);
+
+    return formattedSignatures;
+  } catch (error) {
+    console.error(`Problem found: ${error}`);
```
stanleyyconsensys, line 504: use logger.error if the try catch has to exist

packages/starknet-snap/index.html
```diff
@@ -255,6 +255,16 @@ <h1>Hello, Snaps!</h1>
         <input type="submit" id="signTypedDataMessage" value="Sign Message" />
       </fieldset>
     </form>
+    <form id="signTransaction">
+      <fieldset>
+        <legend>Sign Transaction</legend>
+        <label for="transactions">Transactions</label>
+        <input type="text" id="transactions" name="transactions" /><br />
```
stanleyyconsensys, line 262: add `signTransaction_` as prefix to id and name  of those input field and label (except button) , to avoid any duplication

stanleyyconsensys (changes requested):

packages/starknet-snap/index.html
wantedsystem: what do you suggest here ?

packages/starknet-snap/index.html
wantedsystem: there is no duplications here

stanleyyconsensys: yes, but just to skip the duplication if there is any new method added

packages/starknet-snap/package.json
```diff
@@ -16,7 +16,7 @@
     "watch": "yarn build && concurrently --kill-others --names SNAP,SERVE -c bgMagenta.bold,bgGreen.bold \"yarn run watch:snap\" \"yarn run serve\"",
     "watch:snap": "mm-snap watch",
     "prettier": "prettier --write \"src/**/*.ts\" \"test/**/*.ts\"",
-    "lint": "eslint . --max-warnings 0 -f json -o eslint-report.json",
+    "lint": "eslint . --max-warnings 100 -f json -o eslint-report.json",
```
stanleyyconsensys, line 19: what is the reason to change the max warning?

stanleyyconsensys (approved):

wantedsystem (commented):
LGTM

PR #159, opened 2023-11-02
feat: add declare method

stanleyyconsensys:
- add contract declare method `starkNet_declareContract ` to snap
- allow to review transaction by using snap dialog with copyable txt on `senderAddress`, `contractPayload `, `invocationsDetails` (invocationsDetails will decouple into nonce, maxFee, version to display)
- method input interface will be extend from starknet.js  method `execute`
- method output interface will be extend from starknet.js  method `execute`


packages/starknet-snap/src/utils/snapUtils.ts
```diff
@@ -220,6 +220,57 @@ export function getSigningTxnText(
...
 }
 
+export function getDeclareSnapTxt(
+  senderAddress: string,
+  network: Network,
+  contractPayload: DeclareContractPayload,
+  invocationsDetails?: InvocationsDetails,
+) {
+  const components = [];
+  components.push(text('**Network:**'));
```
wantedsystem, line 230: Maybe add a function to avoid repeating the code : 
```
function addComponent(components, label, value) {
  components.push(text(label));
  components.push(copyable(value));
}
addComponent(components, '**Network:**', network.name);

stanleyyconsensys: will be update for all after those new method PR merge

wantedsystem (approved):

packages/starknet-snap/src/utils/snapUtils.ts
wantedsystem: LGTM, just need to fix conflict :)

PR #161, opened 2023-11-06
feat: add multi transaction to snap

stanleyyconsensys:
- add multi transaction method `starknet_executeTxn` to snap
- allow to review transaction by using snap dialog with copyable txt on `senderAddress`, `txnInvocation`, `abis`, `invocationsDetails` (invocationsDetails will decouple into nonce, maxFee, version to display)
- method input interface will be extend from starknet.js  method `execute`
- method output interface will be extend from starknet.js  method `execute`
- add lint:fix option


wantedsystem (dismissed):
LGTM ;)

PR #162, opened 2023-11-06
feat:  estimate fees

stanleyyconsensys:
- add multi estimate fee method `starknet_estimates` to snap
- method input interface will be extend from starknet.js  method `estimateFeeBulk`
- method output interface will be extend from starknet.js  method `estimateFeeBulk`, but convert value overall_fee, gas_consumed, gas_price, suggestedMaxFee to string to fullfill snap standard
- unit test included


wantedsystem (approved):
LGTM as well just fix the conflict

PR #179, opened 2023-11-22
fix: fix wallet cant connect after disconnect

stanleyyconsensys:
**Description**:
The snap not able to reconnect to the wallet ui after disconnected 
Issue: https://consensyssoftware.atlassian.net/browse/SF-573

**Resolution**:
the root cause is the "ethereum" provider reset after disconnected, and there is no operation to re-init the provider, hence the screen freeze when try to use the provider to request snap connect

to address the problem, not to reset the provider even disconnect, as the provider is not part of the snap wallet, it is the permanent connection with metamask 

**Limitation**:
a user disconnect the app and remove/disable the metamask, and try to click the re-connect button will not take effect

### DIFF
--- a/packages/wallet-ui/src/slices/walletSlice.ts
+++ b/packages/wallet-ui/src/slices/walletSlice.ts
@@ -87,9 +87,10 @@ export const walletSlice = createSlice({
     clearAccounts: (state) => {
       state.accounts = [];
     },
-    resetWallet: () => {
+    resetWallet: (state) => {
       return {
         ...initialState,
+        provider: state.provider,
         forceReconnect: true,
       };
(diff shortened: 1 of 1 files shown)


wantedsystem (approved):

atchoum6760:
Hello I have a problem connecting to Starknet Snap I got a message to update but now when I want to sign with metamask I get the following message any idea how to rectify?    Thanks                                                                                                       Connection failed
The extraction of npm:@consensys/starknet-snap failed, please check your network connection and try again.
Failed to fetch snap "npm:@consensys/starknet-snap": No matching versions of the snap are on the allowlist

PR #184, opened 2023-12-08
feat: display Stark name in Sidebar

### LINKED ISSUE #183: feat: display user stark domain in sidebar
This feature will allow users to see their main domain instead of their hex address. 

Capture d’écran 2023-11-30 à 14 48 46

**Implementation proposal**

To retrieve the stark name from the user address, we can call the function `address_to_domain` from the starknet ID [naming contract](https://starkscan.co/contract/0x6ac597f8116f886fa1c97a23fa4e08299975ecaf6b598873ca6792b9bbfb678#read-write-contract). This function returns an array of felt that need to be decoded to be readable. You can find more information on our encoding and decoding algorithm [here](https://docs.starknet.id/architecture/naming/encoding). 

```
callContract({
    contractAddress: namingContract,
    entrypoint: 'address_to_domain',
    calldata: [address]
});
```

[Source code](https://github.com/starknet-id/naming/blob/master/src/naming/main.cairo#L208)

**In starknet-snap:**

- Create a new method in [Starknet MetaMask Snap API](https://github.com/Consensys/starknet-snap/blob/main/packages/starknet-snap/openrpc/starknet_snap_api_openrpc.json) : `starkNet_getStarkName` 
- In [starknet-snap/src](https://github.com/Consensys/starknet-snap/tree/main/packages/starknet-snap/src) create a `getStarkName` function 

```ts
export async function getStarkName(params: ApiParams) {
  try {
    const { state, requestParams } = params;
    const requestParamsObj = requestParams as GetStarkNameRequestParam;

    // validate requestParamsObj.address argument
    // get network

    // Call a util function to get the stark domain
    const name = await getStarkNameUtil(network, requestParamsObj.address);
    logger.log(`getStarkName: name:\n${toJson(name)}`);

    return name;
  } catch (err) {
    logger.error(`Problem found: ${err}`);
    throw err;
  }
}
```
- In `starknetUtils.ts,` create a util function `getStarkName` that will use the [starknet.js](https://github.com/starknet-io/starknet.js/blob/develop/src/provider/starknetId.ts#L6) function `getStarkName` which already implement calling the starknet ID naming contract and decoding stark domain.

```ts
export const getStarkName = async (
  network: Network,
  address: string,
): Promise<string> => {
  const provider = getProvider(network);
  return provider.getStarkName(address)
};
```

**In wallet-ui**

- In [useStarkNetSnap.ts](https://github.com/Consensys/starknet-snap/blob/main/packages/wallet-ui/src/services/useStarkNetSnap.ts), add a call to our function `starkNet_getStarkName`
- Then we can use it in the [SideBar](https://github.com/Consensys/starknet-snap/blob/main/packages/wallet-ui/src/components/ui/organism/SideBar/SideBar.view.tsx) component
- update [AccountAddress.view.tsx](https://github.com/Consensys/starknet-snap/blob/main/packages/wallet-ui/src/components/ui/molecule/AccountAddress/AccountAddress.view.tsx) to have a new variable `name`

```ts
interface Props {
  address: string;
  full?: boolean;
  placement?: PopperPlacementType;
  // new variable 
  name?: string
}
```

- if `name` is defined then show `name` instead of `address`. It would require to create a `shortenName` function to shorten the name if it is too long to fit in the UI. 

If a significant number of contributors agree with our approach I'll start working on Pull Request.

Please feel free to answer on this issue and tell me what you think !

irisdv:
Close #183 

This PR displays the user Stark name in the `Sidebar` component.

To test it, you can use the python script `buy_domain` from this repo : https://github.com/irisdv/test_scripts

### DIFF
--- a/packages/starknet-snap/src/types/snapApi.ts
+++ b/packages/starknet-snap/src/types/snapApi.ts
@@ -206,3 +206,7 @@ export interface SignDeclareTransactionRequestParams extends Autherizeable, Sign
 export interface SwitchNetworkRequestParams extends Autherizeable, BaseRequestParams {
   chainId: string;
 }
+
+export interface GetStarkNameRequestParam extends BaseRequestParams {
+  userAddress: string;
+}
--- a/packages/starknet-snap/src/utils/starknetUtils.ts
+++ b/packages/starknet-snap/src/utils/starknetUtils.ts
@@ -544,3 +544,8 @@ export const signMessage = async (
   const signatures = await signer.signMessage(typedDataMessage, signerUserAddress);
   return stark.signatureToDecimalArray(signatures);
 };
+
+export const getStarkNameUtil = async (network: Network, userAddress: string) => {
+  const provider = getProvider(network);
+  return provider.getStarkName(userAddress);
+};
--- a/packages/wallet-ui/src/utils/utils.ts
+++ b/packages/wallet-ui/src/utils/utils.ts
@@ -120,3 +120,15 @@ export const fetchWithTimeout = async (resource: string, options = { timeout: TI
   clearTimeout(id);
   return response;
 };
+
+export const shortenDomain = (domain: string, maxLength = 18) => {
+  if (!domain) return '';
+  const ellipsis = '...';
+
+  if (domain.length <= maxLength) {
+    return domain;
+  }
+
+  const shortenedPartLength = maxLength - ellipsis.length;
+  return `${domain.substring(0, shortenedPartLength)}${ellipsis}`;
+};
--- a/packages/wallet-ui/src/utils/utils.ts
+++ b/packages/wallet-ui/src/utils/utils.ts
@@ -4,6 +4,8 @@ import {
   DECIMALS_DISPLAYED_MAX_LENGTH,
   STARKNET_MAINNET_EXPLORER,
   STARKNET_TESTNET_EXPLORER,
+  STARKNET_SEPOLIA_TESTNET_EXPLORER,
+  SEPOLIA_CHAINID,
   TIMEOUT_DURATION,
 } from './constants';
 import { Erc20Token, Erc20TokenBalance } from 'types';
@@ -26,6 +28,9 @@ export const openExplorerTab = (
       break;
     case constants.StarknetChainId.TESTNET:
       explorerUrl = STARKNET_TESTNET_EXPLORER;
+      break;
+    case SEPOLIA_CHAINID:
+      explorerUrl = STARKNET_SEPOLIA_TESTNET_EXPLORER;
       break;
   }
   window.open(explorerUrl + type + '/' + address, '_blank')?.focus();
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,4 +1,4 @@
 {
-  "packages/starknet-snap": "2.3.0",
-  "packages/wallet-ui": "1.18.0"
+  "packages/starknet-snap": "2.4.0",
+  "packages/wallet-ui": "1.19.0"
 }
--- a/packages/starknet-snap/CHANGELOG.md
+++ b/packages/starknet-snap/CHANGELOG.md
@@ -1,6 +1,14 @@
 # Changelog
 
 
+## [2.4.0](https://github.com/Consensys/starknet-snap/compare/starknet-snap-v2.3.0...starknet-snap-v2.4.0) (2023-12-19)
+
+
+### Features
+
+* implement network switch in UI ([#175](https://github.com/Consensys/starknet-snap/issues/175)) ([4b4ace7](https://github.com/Consensys/starknet-snap/commit/4b4ace7f41998c36c7924dcfb07dde061a714d45))
+* sepolia network ([#187](https://github.com/Consensys/starknet-snap/issues/187)) ([45b0463](https://github.com/Consensys/starknet-snap/commit/45b04633149bf1b9568feeb8c2f475d52a0e45a8))
+
 ## [2.3.0](https://github.com/Consensys/starknet-snap/compare/starknet-snap-v2.2.0...starknet-snap-v2.3.0) (2023-11-20)
 
 
--- a/packages/starknet-snap/openrpc/starknet_snap_api_openrpc.json
+++ b/packages/starknet-snap/openrpc/starknet_snap_api_openrpc.json
@@ -1167,6 +1167,40 @@
         }
       },
       "errors": []
+    },
+    {
+      "name": "starkNet_getStarkName",
+      "summary": "Get Stark name from a Starknet account address",
+      "paramStructure": "by-name",
+      "params": [
+        {
+          "name": "userAddress",
+          "summary": "Address of the account contract",
+          "description": "Address of the account contract",
+          "required": true,
+          "schema": {
+            "$ref": "#/components/schemas/ADDRESS"
+          }
+        },
+        {
+          "name": "chainId",
+          "summary": "Id of the target Starknet network",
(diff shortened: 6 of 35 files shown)


packages/wallet-ui/src/utils/utils.ts
```diff
@@ -125,3 +125,15 @@ export const fetchWithTimeout = async (resource: string, options = { timeout: TI
   clearTimeout(id);
   return response;
 };
+
+export const shortenDomain = (domain: string, maxLength = 18) => {
+  if (!domain) return '';
+  const ellipsis = '...';
+
+  if (domain.length <= maxLength) {
```
stanleyyconsensys, line 133: does the domain always in eng? if not we may consider to verify the length by [...domain].length

packages/starknet-snap/src/utils/starknetUtils.ts
```diff
@@ -544,3 +544,8 @@ export const signMessage = async (
   const signatures = await signer.signMessage(typedDataMessage, signerUserAddress);
   return stark.signatureToDecimalArray(signatures);
 };
+
+export const getStarkNameUtil = async (network: Network, userAddress: string) => {
+  const provider = getProvider(network);
+  return provider.getStarkName(userAddress);
+};
```
stanleyyconsensys, line 551: Add EOS

```suggestion
};

```

packages/starknet-snap/src/types/snapApi.ts
```diff
@@ -206,3 +206,7 @@ export interface SignDeclareTransactionRequestParams extends Autherizeable, Sign
 export interface SwitchNetworkRequestParams extends Autherizeable, BaseRequestParams {
   chainId: string;
 }
+
+export interface GetStarkNameRequestParam extends BaseRequestParams {
+  userAddress: string;
+}
```
stanleyyconsensys, line 212: Add EOS

```suggestion
}

```

packages/starknet-snap/src/utils/starknetUtils.ts
```diff
@@ -544,3 +544,8 @@ export const signMessage = async (
   const signatures = await signer.signMessage(typedDataMessage, signerUserAddress);
   return stark.signatureToDecimalArray(signatures);
 };
+
+export const getStarkNameUtil = async (network: Network, userAddress: string) => {
+  const provider = getProvider(network);
+  return provider.getStarkName(userAddress);
```
stanleyyconsensys, line 550: please check if the StarknetIdContract in sepolia
if yes, please pass in the sepolia StarknetIdContract when using sepolia

the current starknet.js does not contains  sepolia contract in getStarknetIdContract

stanleyyconsensys (changes requested):
LGTM, thanks for the work

IMHO, i think there is no reason to create another snap method to get the name individually

instead the logic should bundle together in current recoverAccounts

packages/starknet-snap/src/utils/starknetUtils.ts
irisdv: It's not yet deployed, but it will be in the coming days. We plan on updating starknet.js with it as well

packages/wallet-ui/src/utils/utils.ts
irisdv: For now yes, but it could change in the future so I updated this part

irisdv:
> LGTM, thanks for the work
> 
> IMHO, i think there is no reason to create another snap method to get the name individually
> 
> instead the logic should bundle together in current recoverAccounts

I updated the code to get the stark name in `recoverAccounts`. I'm just not entirely sure of the best way to store it to use it in the ui, for now I created a `state.starkNames` but maybe storing it in `state.accounts` would make more sense ?

stanleyyconsensys:
> > LGTM, thanks for the work
> > IMHO, i think there is no reason to create another snap method to get the name individually
> > instead the logic should bundle together in current recoverAccounts
> 
> I updated the code to get the stark name in `recoverAccounts`. I'm just not entirely sure of the best way to store it to use it in the ui, for now I created a `state.starkNames` but maybe storing it in `state.accounts` would make more sense ?

Thank you very much for the work

For SNAP
in order to bundle with recoverAccounts, you may have also need to update the methods of `upsertAccount` in snapUtils

For UI
you may need to change the structure of setAccounts , instead of storing a string address, but a object {address, name}
you may need to change the all files that need to using the address from state

in such big changes, may be it is better to rollback to use your original implementation, a single snap api to get the starkname

sorry for causing you the inconvenient

irisdv:
> > > LGTM, thanks for the work
> > > IMHO, i think there is no reason to create another snap method to get the name individually
> > > instead the logic should bundle together in current recoverAccounts
> > 
> > 
> > I updated the code to get the stark name in `recoverAccounts`. I'm just not entirely sure of the best way to store it to use it in the ui, for now I created a `state.starkNames` but maybe storing it in `state.accounts` would make more sense ?
> 
> Thank you very much for the work
> 
> For SNAP in order to bundle with recoverAccounts, you may have also need to update the methods of `upsertAccount` in snapUtils
> 
> For UI you may need to change the structure of setAccounts , instead of storing a string address, but a object {address, name} you may need to change the all files that need to using the address from state
> 
> in such big changes, may be it is better to rollback to use your original implementation, a single snap api to get the starkname
> 
> sorry for causing you the inconvenient

No worries, I rolled back to the previous version !

stanleyyconsensys:
> > > > LGTM, thanks for the work
> > > > IMHO, i think there is no reason to create another snap method to get the name individually
> > > > instead the logic should bundle together in current recoverAccounts
> > > 
> > > 
> > > I updated the code to get the stark name in `recoverAccounts`. I'm just not entirely sure of the best way to store it to use it in the ui, for now I created a `state.starkNames` but maybe storing it in `state.accounts` would make more sense ?
> > 
> > 
> > Thank you very much for the work
> > For SNAP in order to bundle with recoverAccounts, you may have also need to update the methods of `upsertAccount` in snapUtils
> > For UI you may need to change the structure of setAccounts , instead of storing a string address, but a object {address, name} you may need to change the all files that need to using the address from state
> > in such big changes, may be it is better to rollback to use your original implementation, a single snap api to get the starkname
> > sorry for causing you the inconvenient
> 
> No worries, I rolled back to the previous version !

just keep the domain.length checking where you did
thank you very much

irisdv:
> > > > > LGTM, thanks for the work
> > > > > IMHO, i think there is no reason to create another snap method to get the name individually
> > > > > instead the logic should bundle together in current recoverAccounts
> > > > 
> > > > 
> > > > I updated the code to get the stark name in `recoverAccounts`. I'm just not entirely sure of the best way to store it to use it in the ui, for now I created a `state.starkNames` but maybe storing it in `state.accounts` would make more sense ?
> > > 
> > > 
> > > Thank you very much for the work
> > > For SNAP in order to bundle with recoverAccounts, you may have also need to update the methods of `upsertAccount` in snapUtils
> > > For UI you may need to change the structure of setAccounts , instead of storing a string address, but a object {address, name} you may need to change the all files that need to using the address from state
> > > in such big changes, may be it is better to rollback to use your original implementation, a single snap api to get the starkname
> > > sorry for causing you the inconvenient
> > 
> > 
> > No worries, I rolled back to the previous version !
> 
> just keep the domain.length checking where you did thank you very much

done

stanleyyconsensys:
LGTM

packages/starknet-snap/src/utils/starknetUtils.ts
irisdv: I made a PR to starknet.js adding the Sepolia contracts but it won't be deployed in production before a few weeks. I can add them in this PR, what do you think ?

PR #193, opened 2024-01-15
feat: refactor with account contract class

stanleyyconsensys:
Refactor with account contract class

- a new class instance `AccountContractService` provide the functionality to get  correct `AccountContract` instance by version and deploy status
- a new class instance `AccountContract` provide the isolation and diversity for different cairo version, such as upgrade, get version, get public key - which may different across different cairo version contract
- a new class instance `AccountKeyring` provide the interface for snap to generate new account and get account contract by address

apply those new class instance into recover account and upgrade account method for test run


packages/starknet-snap/src/services/account-contract/contract.ts
```diff
@@ -0,0 +1,61 @@
...
+    const accountContracts: Array<AccountContract> = [];
+    for (let i = 0; i < this.#accountContracts.length; i++) {
+      const account = this.#accountContracts[i];
+      const accountContract = account.FromSeed(seed, this.provider);
+      accountContracts.push(accountContract);
+    }
+    return accountContracts;
+  }
+
+  async isContractUpgraded(contract: AccountContract, reflesh = false): Promise<boolean> {
```
wantedsystem, line 54: You mean `refresh` instead of `reflesh` ? if yes we need to modify it in all the other files

packages/starknet-snap/src/services/account-contract/contracts/base.ts
```diff
@@ -0,0 +1,110 @@
...
+        await this.getVersion();
+        this.deployed = true;
+      } catch (err) {
+        if (!err.message.includes('Contract not found')) {
+          throw err;
+        }
+        this.deployed = false;
+      }
+    } else {
+      console.log("cached")
```
wantedsystem, line 73: Better to use logger.log here ?

packages/starknet-snap/src/services/account-contract/contracts/cairo0.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export class CairoZeroContract
+  extends AccountContractBase
+  implements StaticImplements<AccountContractStatic, typeof CairoZeroContract>
+{
+  static ClassHash: string = ACCOUNT_CLASS_HASH_LEGACY;
+  static ProxyClassHash: string = PROXY_CONTRACT_HASH;
+  static CairoVersion = '0' as CairoVersion;
+
+  constructor(address: string, pubKey: string, priKey: string, provider?: NodeProvider) {
```
wantedsystem, line 19: To maintain consistency could we use `prvKey` instead of `priKey` ?

packages/starknet-snap/src/services/account-contract/contracts/cairo0.ts
```diff
@@ -0,0 +1,76 @@
...
+      0,
+    );
+    return padAddress(address);
+  }
+
+  static GetCallData(pubKey: string): Calldata {
+    return CallData.compile({
+      implementation: CairoZeroContract.ClassHash,
+      selector: hash.getSelectorFromName('initialize'),
+      calldata: CallData.compile({ signer: pubKey, guardian: '0' }),
```
wantedsystem, line 37: Can we put `guardian: '0'` in constants.ts ?

packages/starknet-snap/src/services/account-contract/contracts/cairo1.ts
```diff
@@ -0,0 +1,56 @@
...
+import { padAddress } from '../../../utils/formatterUtils';
+
+export class CairoOneContract
+  extends AccountContractBase
+  implements StaticImplements<AccountContractStatic, typeof CairoOneContract>
+{
+  static ClassHash: string = ACCOUNT_CLASS_HASH;
+  static CairoVersion = '1' as CairoVersion;
+
+  constructor(address: string, pubKey: string, priKey: string, provider?: NodeProvider) {
```
wantedsystem, line 17: same here for `prvKey` instead of `priKey`

packages/starknet-snap/src/services/account-contract/contracts/cairo1.ts
```diff
@@ -0,0 +1,56 @@
...
+      CairoOneContract.GetCallData(pubKey),
+      0,
+    );
+    return padAddress(address);
+  }
+
+  static GetCallData(pubKey: string): Calldata {
+    return CallData.compile({
+      signer: pubKey,
+      guardian: '0',
```
wantedsystem, line 34: same here for guardian

packages/starknet-snap/src/services/account/snapState.ts
```diff
@@ -0,0 +1,92 @@
...
+  async saveMany(userAccounts: Array<AccContract>, chainId:string): Promise<void> {
+    return this.lock.runExclusive(async () => {
+      const state = await this.getState();
+      const accounts = await this.getAccounts(chainId);
+      const accountMap = new Map(accounts.map((acc) => [acc.address, acc]));
+      if (!state.accContracts) {
+        state.accContracts = [];
+      }
+
+      for (var i = 0; i< userAccounts.length; i++) {
```
wantedsystem, line 71: I think it's better to use `let i = 0` instead of `var i = 0`

packages/starknet-snap/src/services/node/provider.ts
```diff
@@ -0,0 +1,53 @@
+import { Provider, RawCalldata, CallContractResponse, ProviderOptions } from 'starknet';
+
+import { Network } from '../../types/snapState';
+
+export interface NodeProvider {
```
wantedsystem, line 5: I think it's better to change the name of the interface because it's the same as the class here

stanleyyconsensys: i will keep this name until i found 1 better, and it is okay in JS for the interface has same name with class

PR #234, opened 2024-04-17
fix: address field error

stanleyyconsensys:
This PR is a temp fix for replacing `address` field by `text` field  due to metamask latest version bug.

### DIFF
--- a/packages/starknet-snap/src/index.ts
+++ b/packages/starknet-snap/src/index.ts
@@ -49,7 +49,7 @@ import { logger } from './utils/logger';
 import { getStarkName } from './getStarkName';
 
 import type { OnRpcRequestHandler, OnHomePageHandler, OnInstallHandler, OnUpdateHandler } from '@metamask/snaps-sdk';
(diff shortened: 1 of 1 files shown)


packages/starknet-snap/src/index.ts
```diff
@@ -288,7 +288,7 @@ export const onHomePage: OnHomePageHandler = async () => {
       const userAddress = accContract.address;
       const chainId = accContract.chainId;
       const network = getNetworkFromChainId(state, chainId);
-      panelItems.push(row('Address', address(`${userAddress}`)));
+      panelItems.push(row('Address', text(`${userAddress}`)));
```
Montoya, line 291: I think you should import copyable and then do two lines: 

```
panelItems.push(text('Address')); 
panelItems.push(copyable(`${userAddress}`)); 
```

stanleyyconsensys: updated

mario-christopher (approved):

Montoya (approved):

PR #243, opened 2024-05-23
feat: add webpack.config.dev.js for local serving of remoteEntry.js

khanti42:
Local serving of remoteEntry.js can be done with 

`yarn serve:fed` from the `packages/get-starknet` sub folder. 
Associated with https://github.com/Consensys/get-starknet/pull/13


packages/get-starknet/webpack.config.dev.js
```diff
@@ -0,0 +1,15 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const { merge } = require('webpack-merge');
+const common = require('./webpack.common.js');
```
stanleyyconsensys, line 3: add // eslint-disable-next-line @typescript-eslint/no-var-requires for each require

stanleyyconsensys (changes requested):
if we wanna setup a complete local dev env
we also need to make this get-starknet wrapper using a SNAP from local rather from prod

in order to do that
update `MetaMaskSnapWallet` class to change the const SNAPI_ID reading from env var , default is `npm:@consensys/starknet-snap`

file: wallet.ts
```javascript
private static readonly SNAPI_ID =  process.env.SNAP_ID ?? npm:@consensys/starknet-snap
```

file: .env
```bash
SNAPI_ID=local:http://localhost:8081
```

stanleyyconsensys (approved):
LGTM

khanti42:
Check  [PR-249](https://github.com/Consensys/starknet-snap/pull/249) for update.

PR #247, opened 2024-06-06
feat: show upgrade dialog on get-starknet calls

khanti42:
This PR is to add upgrade dialog on below APIs where those are using by get-starknet:

- `starkNet_signDeclareTransaction`
- `starkNet_signDeployAccountTransaction`
- `starkNet_executeTxn`
- `starkNet_declareContract`

if above APIS trigger with an Cairo 0, account contract, it will first display an dialog, then reject the request by throw an Error

image


packages/starknet-snap/src/declareContract.ts
```diff
@@ -16,6 +16,11 @@ export async function declareContract(params: ApiParams) {
     const senderAddress = requestParamsObj.senderAddress;
     const network = getNetworkFromChainId(state, requestParamsObj.chainId);
     const { privateKey } = await getKeysFromAddress(keyDeriver, network, state, senderAddress);
+    
+    if(isUpgradeRequired(network, senderAddress)){
+      showUpgradeRequestModal(wallet);
+      return false;
```
stanleyyconsensys, line 22: perhaps throw an error is better?
in terms of end user exp, we handle by dialog
in terms of the coding handle, return false may not fit to the purpose for the caller to determine it is success or not , throw an error seem more general 

WDYT

stanleyyconsensys (changes requested):

packages/starknet-snap/src/declareContract.ts
stanleyyconsensys: BTW , test is not pass

stanleyyconsensys: you have to await for showUpgradeRequestModal

khanti42:
I already merged latest changes from main (including the deployment check for executeTxn), the PR#254 including these changes in release/2.3.0 is still open

stanleyyconsensys (approved):
LGTM

stanleyyconsensys (approved):

PR #249, opened 2024-06-06
feat: Enable Local Testing of Webpack Federation Module (get-starknet remoteEntry.js)

khanti42:


packages/get-starknet/src/wallet.ts
```diff
@@ -27,7 +27,7 @@ export class MetaMaskSnapWallet implements IStarknetWindowObject {
   metamaskProvider: MetaMaskProvider;
 
   private static readonly cairoVersion = '0';
-  private static readonly SNAPI_ID = 'npm:@consensys/starknet-snap';
+  private static readonly SNAPI_ID =  process.env.SNAP_ID;
```
stanleyyconsensys, line 30: i think better to have default value, as the CICD doesnt have it

stanleyyconsensys (approved):

PR #251, opened 2024-06-07
feat: add waiting mode for create account

stanleyyconsensys:
Due to a issue of [SF-636](https://consensyssoftware.atlassian.net/jira/software/projects/SF/boards/472?selectedIssue=SF-636)

This PR is to add waitForTransaction on create account rpc method, to allow the snap to wait for the deploy txn successfully land on L2

 

[SF-636]: https://consensyssoftware.atlassian.net/browse/SF-636?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/starknet-snap/src/createAccount.ts
+++ b/packages/starknet-snap/src/createAccount.ts
@@ -6,6 +6,7 @@ import {
   callContract,
   estimateAccountDeployFee,
   getSigner,
+  waitForTransaction,
 } from './utils/starknetUtils';
 import {
   getEtherErc20Token,
@@ -23,7 +24,7 @@ import { DialogType } from '@metamask/rpc-methods';
 import { heading, panel, text } from '@metamask/snaps-sdk';
 import { logger } from './utils/logger';
 
-export async function createAccount(params: ApiParams, silentMode = false) {
+export async function createAccount(params: ApiParams, silentMode = false, waitMode = false) {
   try {
     const { state, wallet, saveMutex, keyDeriver, requestParams } = params;
     const requestParamsObj = requestParams as CreateAccountRequestParams;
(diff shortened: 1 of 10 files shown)


packages/starknet-snap/src/utils/starknetUtils.ts
```diff
@@ -108,6 +110,18 @@ export const estimateFee = async (
...
+export const waitForTransaction = async (
+  network: Network,
+  senderAddress: string,
+  privateKey: string | Uint8Array,
+  txnHash: num.BigNumberish,
+  cairoVersion?: CairoVersion,
+): Promise<GetTransactionReceiptResponse> => {
+  const provider = getProvider(network);
+  const account = new Account(provider, senderAddress, privateKey, cairoVersion ?? '0');
+  return account.waitForTransaction(txnHash);
```
mario-christopher, line 122: Should you `await` here ?

packages/starknet-snap/src/createAccount.ts
```diff
@@ -148,6 +149,10 @@ export async function createAccount(params: ApiParams, silentMode = false) {
 
       logger.log(`createAccount:\ndeployResp: ${toJson(deployResp)}`);
 
+      if (waitMode) {
```
mario-christopher, line 152: Can you add a comment here describing when we would `wait` and when we would not ?

packages/starknet-snap/src/utils/starknetUtils.ts
stanleyyconsensys: in terms of the functionality, it is not required, as the caller has to use await, 
or
we can add this to the rest of the same pattern function in 1 batch to standardise the pattern

packages/starknet-snap/src/createAccount.ts
stanleyyconsensys: sure, the wait mode basically is only use when you need other txn compose together, and to make sure this txn execute complete, the wait mode is something that we need

stanleyyconsensys: added

khanti42 (approved):
LGTM!

PR #255, opened 2024-06-18
feat: sf-640 revamp cicd workflow

stanleyyconsensys:
This PR is to clean up the CICD and try to standardise with metamask workflow if possible

- add get starknet deployment flow
- remove story book workflow, as we dont have any use case for now
- combine deploy-dev, publish-npm-dev into deploy.yml, to let github taking care the dynamic variable "AWS_S3_URL" and "AWS_CLOUDFRONT_DISTRIBUTIONS_ID", and simplify the workflow
- standardise package.json cmd, to have  start, build, test, lint, lint, clean, serve on every workspace
- remove yarn setup, as it is duplicate with yarn install, instead replace by allow-scripts from root
- remove e2e, as it is not enable for now

**New workflow:**
Main (trigger by any push or pull request to main, to test workflow and the code)
image

Deploy (trigger by manual, allow deploy to dev,staging,prod)
image

**Existing workflow:**
lint-pr-title, terraform, release-please will remain the same

**Removed workflow:**
e2e-tests -> removed with no use case
e2e-tests-slack-notification -> removed with no use case
storybook -> removed with no use case
deploy-dev -> merged to deploy
publish-npm-dev -> merged to deploy
build-test -> merged to main

Note:
In MM workflow, they dont have dev , staging deployment, but mostly 1 go to PROD by manual create a release PR request, and merge the PR, and then the pipeline will deploy to prod
in Starknet workflow, we do it by auto create release PR when ever a push to main, and manual deploy to different env for more flexible choice, e.g we can deploy a version b4 release PR or after release PR

Ticket: 
https://consensyssoftware.atlassian.net/jira/software/projects/SF/boards/472?selectedIssue=SF-640


.github/workflows/deploy.yml
```diff
@@ -8,84 +8,202 @@ on:
...
         required: true
 jobs:
-  deploy:
+  perpare-deployment:
     environment: ${{ inputs.environment }}
-    name: Deploy Snap and UI
-    runs-on: ubuntu-20.04
+    permissions:
+      contents: write
+    runs-on: ubuntu-latest
```
khanti42, line 15: should we fix the version instead of using ubuntu-latest to control the image version we use ?

packages/wallet-ui/package.json
```diff
@@ -70,21 +70,19 @@
...
-    "lint:react": "eslint ./src/**/*.ts ./src/**/*.tsx -f json -o eslint-report.json",
-    "lint:misc": "prettier --check ./src/**/*.js ./src/**/*.ts ./src/**/*.tsx",
-    "pretty": "prettier --write ./src/**/*.js  ./src/**/*.ts ./src/**/*.tsx",
-    "lint": "yarn lint:react && yarn lint:misc",
-    "clean": "rimraf build",
     "storybook": "start-storybook -p 6006",
-    "build-storybook": "build-storybook",
-    "type:check": "tsc --pretty --noEmit",
-    "release": "sh ../../scripts/release.sh"
+    "build-storybook": "build-storybook"
```
khanti42, line 85: do we still need this ?

khanti42 (changes requested):
LGTM. Just small comment regarding ubuntu image version and the storybook entry. Otherwise do you think it makes sense to add the automatic deployment of remoteEntry.js to s3 into the cicd workflow already with this merge ?

packages/wallet-ui/package.json
stanleyyconsensys: good catch, we can remove, but lets do it on other PR, i have a PR to clean up the package.json as well

.github/workflows/deploy.yml
stanleyyconsensys: may be fixing a version is better, lets move it to static version then

stanleyyconsensys: i just find that it is using ubuntu-latest everywhere, and only this doc is using ubuntu-20.04 
so may be to keep it safe , use ubuntu-latest?

khanti42: ok

khanti42 (approved):
LGTM

PR #256, opened 2024-06-18
fix: sf 639 update contract upgrade message

stanleyyconsensys:
This PR is to fix the update contract upgrade message

from 
`Waiting for transaction finalize... `

to
`Waiting for transaction to be finalised.`


khanti42 (dismissed):
LGTM

PR #257, opened 2024-06-19
fix: allow immutable yarn lock

khanti42:
The current CI/CD uses --no-immutable option which is discouraged. The goal of this PR is to remove the need for this option by changing the package.json of the wallet-ui package and change the CI/CD pipelines accordingly. 

Jira: https://consensyssoftware.atlassian.net/jira/software/projects/SF/boards/472?selectedIssue=SF-645


stanleyyconsensys (changes requested):
it doesnt seem work , it is still depends on @consensys/starknet-snap from the code base

PR #258, opened 2024-06-20
fix: resolve issue in cicd pipeline

stanleyyconsensys:
This PR is to fix the missing cache in cicd deploy pipeline


.github/workflows/deploy.yml
```diff
@@ -190,15 +211,19 @@ jobs:
...
+            ./packages/starknet-snap/snap.manifest.json
             ./node_modules/.yarn-state.yml
           key: ${{ github.sha }}
       - name: Deploy to AWS
         run: | 
           echo "Deployed Dapp to : $AWS_S3_URL"
           aws s3 sync ./packages/wallet-ui/build "$AWS_S3_URL"
         env:
-          AWS_S3_URL: ${{ vars.AWS_S3_URL }}
+          AWS_S3_URL: ${{ needs.perpare-deployment.outputs.AWS_S3_URL }}
```
khanti42, line 226: why do we use needs.perpare-deployment.outputs instead of vars here ? Also there might be a typo on needs.perpare maybe it should be needs.prepare ?

khanti42 (changes requested):
Maybe just check the naming of the needs.perpare --~ needs.prepare. 

Otherwise LGTM

.github/workflows/deploy.yml
stanleyyconsensys: yup , i will fix the typo: needs.prepare

and yes , using vars u have to take an extra AGR on the workflow, telling the workflow to extra param from action parameter store

hence, i use the prepare parameter workflow to do everything, and output the required var

and the following workflow will just need to use needs.perpare-deployment to get the param

khanti42 (dismissed):
LGTM

khanti42 (approved):
LGTM

PR #259, opened 2024-06-21
feat: clean up and update dependency

stanleyyconsensys:
This PR is to standardize the package.json on each project to be the same / similar as Metamask's Snap

What did we updated:
- update npmrc to use lts/* rather than hard code as 16 
- Not necessary to using patch version of yarn (yarn-berry), instead we can use yarn 3.2.1
- remove metamask/rpc-method, as it has migrate into metamask/sdk
- move allows-script cmd to root level for cleaning code, as it seem redundant 
- remove release.sh , as we cant release from local anyway
- remove some un-used dependencies


khanti42:
Still need to check workflow with get-starknet.

stanleyyconsensys:
@khanti42 verified the get-starknet is working too

packages/starknet-snap/package.json
```diff
@@ -1,39 +1,51 @@
...
     "clean": "rimraf dist",
-    "start": "mm-snap watch",
-    "test": "yarn run test:unit && yarn run cover:report",
+    "cover:report": "nyc report --reporter=lcov --reporter=text",
     "lint": "yarn lint:eslint && yarn lint:misc --check",
     "lint:eslint": "eslint . --cache --ext js,ts",
     "lint:fix": "yarn lint:eslint --fix && yarn lint:misc --write",
     "lint:misc": "prettier '**/*.ts' '**/*.json' '**/*.md' '!CHANGELOG.md' --ignore-path .gitignore",
-    "test:unit": "nyc --check-coverage --statements 80 --branches 80 --functions 80 --lines 80 mocha --colors -r ts-node/register \"test/**/*.test.ts\"",
-    "test:unit:one": "nyc --check-coverage --statements 80 --branches 80 --functions 80 --lines 80 mocha --colors -r ts-node/register",
```
khanti42, line 27: can we keep this test:unit:one ? It's handy when working on one test file to not have to run all the tests every time.

khanti42 (changes requested):
LGTM!

Just the small comment on unit testing single file. But this is not very important.

packages/starknet-snap/package.json
stanleyyconsensys: sure

stanleyyconsensys: done

khanti42 (approved):
LGTM!

PR #269, opened 2024-07-01
feat: allow txv3 in get-starknet execute txn

khanti42:
`get-starknet` users (they are dapp dev) interact with the starknet blockchain by calling execute which has the following signature : 

`abstract execute(transactions: AllowArray<Call>, transactionsDetail?: InvocationsDetails): Promise<InvokeFunctionResponse>;`

The InvocationsDetails type defined by `starknet.js` is as follow 

```
type InvocationsDetails = {
    nonce?: BigNumberish;
    maxFee?: BigNumberish;
    version?: BigNumberish;
```

The transaction version can be set there.

This is how we can handle the version on `executeTxn` rpc endpoint used by our federated module middleware.

### DIFF
--- a/packages/starknet-snap/src/executeTxn.ts
+++ b/packages/starknet-snap/src/executeTxn.ts
@@ -1,9 +1,8 @@
 import type { Component } from '@metamask/snaps-sdk';
 import { heading, panel, divider, DialogType } from '@metamask/snaps-sdk';
-import type { Invocations } from 'starknet';
+import type { constants, Invocations } from 'starknet';
 import { TransactionType } from 'starknet';
 
-import { createAccount } from './createAccount';
 import type {
   ApiParamsWithKeyDeriver,
   ExecuteTxnRequestParams,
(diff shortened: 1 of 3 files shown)


stanleyyconsensys:
i think the PR is not ready, shall we move it back to draft

packages/starknet-snap/src/executeTxn.ts
```diff
@@ -133,17 +124,29 @@ export async function executeTxn(params: ApiParamsWithKeyDeriver) {
...
 
     if (!accountDeployed) {
-      await createAccount(createAccountApiParams, true, true);
+      await createAccount({
+        network,
+        address,
+        publicKey,
+        privateKey,
+        waitMode: true,
+        callback: undefined,
```
stanleyyconsensys, line 133: you dont need undefine it, but since this api will drop, let it pass

packages/starknet-snap/src/utils/snapUtils.ts
```diff
@@ -392,9 +392,13 @@ export function getTxnSnapTxt(
   }
 
   if (details?.maxFee) {
+    const feeToken: FeeToken =
```
stanleyyconsensys, line 395: i saw this logic twice, we mat consider to move to a function soon

stanleyyconsensys (approved):
LGTM

PR #270, opened 2024-07-02
fix: add cache key for different env in CICD

stanleyyconsensys:
This PR is to fix a bug in the CICD pipeline when processiong “Deployments" workflow

if the same commit point deploy twice in different env, the latter deploy code will reuse the build from prev build from a cache, and lead to incorrect deployment

solution:
adding the cache with key `commit sha + '-' + env`

jira: https://consensyssoftware.atlassian.net/jira/software/projects/SF/boards/472?selectedIssue=SF-651


khanti42 (approved):
LGTM

PR #272, opened 2024-07-02
fix: remove get balance from createAccount

stanleyyconsensys:
This PR is to fix, the get balance call is fail during create account 

the solution is to remove the get balance call, as it doesnt needed , starknet.js will auto estimate for user, 

Note: it may break tx v3, but we should fix it in tx v3, but not this PR

jira:
https://consensyssoftware.atlassian.net/jira/software/projects/SF/boards/472?selectedIssue=SF-650

### DIFF
--- a/packages/starknet-snap/src/createAccount.ts
+++ b/packages/starknet-snap/src/createAccount.ts
@@ -3,13 +3,9 @@ import {
   getKeysFromAddressIndex,
   getAccContractAddressAndCallData,
   deployAccount,
-  getBalance,
-  estimateAccountDeployFee,
-  isAccountDeployed,
   waitForTransaction,
 } from './utils/starknetUtils';
 import {
-  getEtherErc20Token,
   getNetworkFromChainId,
   getValidNumber,
   upsertAccount,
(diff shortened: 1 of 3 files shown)


packages/starknet-snap/src/createAccount.ts
```diff
@@ -78,47 +70,8 @@ export async function createAccount(params: ApiParams, silentMode = false, waitM
...
-        contractCallData,
-        publicKey,
-        privateKey,
-        undefined,
-        {
-          maxFee: estimateDeployFee?.suggestedMaxFee,
-        },
-      );
+      // Deploy account will auto estimate the fee from the network if not provided
+      const deployResp = await deployAccount(network, contractAddress, contractCallData, publicKey, privateKey);
```
khanti42, line 74: What happens if createAccount is called on an account that is already deployed ?

packages/starknet-snap/src/createAccount.ts
```diff
@@ -51,8 +44,7 @@ export async function createAccount(params: ApiParams, silentMode = false, waitM
       `createAccount:\ncontractAddress = ${contractAddress}\npublicKey = ${publicKey}\naddressIndex = ${addressIndexInUsed}`,
     );
 
-    let failureReason = '';
-    let estimateDeployFee: EstimateFee;
+    const failureReason = '';
```
khanti42, line 47: shall we still use this ? Maybe in case the account was already deployed ?

khanti42 (commented):
LGTM, just small comment regarding the failureReason.

packages/starknet-snap/src/createAccount.ts
stanleyyconsensys: it will fail in this call, which is the same behaviour as prev logic

packages/starknet-snap/src/createAccount.ts
stanleyyconsensys: it will fail in this call, which is the same behaviour as prev logic

stanleyyconsensys: we dont need this, yes, we should remove

Issue #150, opened 2023-09-27
Bug Report: Fail to sign multiple transactions consecutively


ColinWttt:
Bug: When signing multiple transactions in succession manually, the second and subsequent transactions fail after signing. After some time, I am able to successfully sign transactions.

This error can be reproduced on [snaps.consensys.net/starknet](https://snaps.consensys.net/starknet), when you send tokens consecutively, the second one fails.

[![error.png](https://i.postimg.cc/jdLJKZzZ/error.png)](https://postimg.cc/gnb0HytZ)

console log two types of errors randomly
error1: MetaMask - RPC Error: 59: A transaction with the same hash already exists in the mempool

error2: MetaMask - RPC Error: 52: Invalid transaction nonce
[![error52.png](https://i.postimg.cc/fRSX7ByJ/error52.png)](https://postimg.cc/rDcKMGSk)

starknet-snap version: @consensys/starknet-snap 2.2.0

MetaMask version: 11.0.0

Chrome: 117.0.5938.92 (arm 64)

Network: Starknet Goerli Testnet

stanleyyconsensys:
the issue was introduced when the transaction has sent, but it is still pending at the block, hence the latest block nonce is not updated on chain.

when we send another transaction in between the nonce update, it will occur such issue

we can resolve the issue by 
Opt1 : adding a wait for transaction to block the UI 
Opt2 : increment the nonce by adding a memory store to remember the last nonce from snap
Opt3 : retry the request with nonce increment when it has nonce error

japuiglaucia:
Bug Report: Fail to sign multiple transactions consecutively #150

Spunsd:
Hello

vivek-consensys:
Tested the fix by @khanti42 on dev and the user is able to sign multiple transactions consecutively. 
![image](https://github.com/user-attachments/assets/dcdc4c98-716e-4f7e-a3dd-2c5b4c71af15)

PR #289, opened 2024-07-11
fix: allow multiple consecutive transactions in estimateFee(Bulk)

khanti42:
Closes : #150 

this PR fixes the gas fee estimation issue when a transaction is in the pending block.

image


packages/starknet-snap/src/utils/starknetUtils.ts
```diff
@@ -139,11 +139,32 @@ export const estimateFee = async (
   cairoVersion?: CairoVersion,
   invocationsDetails?: UniversalDetails,
 ): Promise<EstimateFee> => {
-  return getAccountInstance(network, senderAddress, privateKey, cairoVersion).estimateInvokeFee(txnInvocation, {
-    ...invocationsDetails,
-    skipValidate: false,
-    blockIdentifier: 'latest',
-  });
+  try {
```
stanleyyconsensys, line 142: we dont need this, as estimateInvokeFee is automatically getting nonce already

packages/starknet-snap/src/estimateFee.ts
```diff
@@ -90,18 +89,9 @@ export async function estimateFee(params: ApiParams) {
...
-      // This condition branch will be removed later when starknet.js
-      // supports estimateFeeBulk in rpc mode
-      estimateFeeResp = await estimateFeeUtil(network, senderAddress, senderPrivateKey, txnInvocation);
-      logger.log(`estimateFee:\nestimateFeeUtil estimateFeeResp: ${toJson(estimateFeeResp)}`);
-    } else {
-      const estimateBulkFeeResp = await estimateFeeBulk(network, senderAddress, senderPrivateKey, bulkTransactions);
-      logger.log(`estimateFee:\nestimateFeeBulk estimateBulkFeeResp: ${toJson(estimateBulkFeeResp)}`);
-      estimateFeeResp = addFeesFromAllTransactions(estimateBulkFeeResp);
-    }
+    const estimateBulkFeeResp = await estimateFeeBulk(network, senderAddress, senderPrivateKey, bulkTransactions);
```
stanleyyconsensys, line 92: the different between estimateFeeBulk and estimateFeeUtil
where estimateFeeBulk doesnt do auto get nonce
but estimateFeeUtil did

but i think it is okay to remote too

packages/starknet-snap/src/utils/starknetUtils.ts
```diff
@@ -154,11 +175,22 @@ export const estimateFeeBulk = async (
   invocationsDetails?: UniversalDetails,
   cairoVersion?: CairoVersion,
 ): Promise<EstimateFee[]> => {
-  return getAccountInstance(network, senderAddress, privateKey, cairoVersion).estimateFeeBulk(txnInvocation, {
-    ...invocationsDetails,
-    skipValidate: false,
-    blockIdentifier: 'latest',
-  });
+  try {
```
stanleyyconsensys, line 178: i think we have to do , as estimateFeeBulk is never getting the nonce

```typescript
if (!invocationsDetails || !invocationsDetails?.nonce) {
  const nonce = await getAccountNonce(network, senderAddress, privateKey);
  if (!invocationsDetails) {
    invocationsDetails = {};
  }
  invocationsDetails.nonce = nonce;
}
```

stanleyyconsensys: and we dont need the try catch

stanleyyconsensys (changes requested):

packages/starknet-snap/src/utils/starknetUtils.ts
```diff
@@ -186,6 +186,21 @@ export const declareContract = async (
   });
 };
 
+export const getAccountNonce = async (
```
stanleyyconsensys, line 189:
```suggestion
export const getLatestAccountNonce = async (
```

packages/starknet-snap/src/utils/starknetUtils.ts
```diff
@@ -194,13 +209,18 @@ export const estimateFee = async (
   cairoVersion?: CairoVersion,
   invocationsDetails?: UniversalDetails,
 ): Promise<EstimateFee> => {
-  return getAccountInstance(
+  let nonce = invocationsDetails?.nonce;
+  if (!nonce) {
+    nonce = await getAccountNonce(network, senderAddress, privateKey);
```
stanleyyconsensys, line 214:
```suggestion
    nonce = await getLatestAccountNonce(network, senderAddress, privateKey);
```

packages/starknet-snap/src/utils/starknetUtils.ts
```diff
@@ -214,13 +234,18 @@ export const estimateFeeBulk = async (
   invocationsDetails?: UniversalDetails,
   cairoVersion?: CairoVersion,
 ): Promise<EstimateFee[]> => {
-  return getAccountInstance(
+  let nonce = invocationsDetails?.nonce;
+  if (!nonce) {
+    nonce = await getAccountNonce(network, senderAddress, privateKey);
```
stanleyyconsensys, line 239:
```suggestion
    nonce = await getLatestAccountNonce(network, senderAddress, privateKey);
```

packages/starknet-snap/src/estimateFee.ts
```diff
@@ -108,35 +107,21 @@ export async function estimateFee(params: ApiParamsWithKeyDeriver) {
...
-        bulkTransactions,
-      );
-      logger.log(
-        `estimateFee:\nestimateFeeBulk estimateBulkFeeResp: ${toJson(
-          estimateBulkFeeResp,
-        )}`,
-      );
-      estimateFeeResp = addFeesFromAllTransactions(estimateBulkFeeResp);
+    if (!estimateFeeResp.suggestedMaxFee || !estimateFeeResp.overall_fee) {
+      throw new Error(`Bulk Fee estimation failed`);
```
stanleyyconsensys, line 124: it looks like we dont have gasConsumed and gasPrice if we use estimateFeeBulk, shall we remove it?

packages/starknet-snap/src/estimateFee.ts
```diff
@@ -108,35 +107,21 @@ export async function estimateFee(params: ApiParamsWithKeyDeriver) {
...
-        senderPrivateKey,
-        bulkTransactions,
-      );
-      logger.log(
-        `estimateFee:\nestimateFeeBulk estimateBulkFeeResp: ${toJson(
-          estimateBulkFeeResp,
-        )}`,
-      );
-      estimateFeeResp = addFeesFromAllTransactions(estimateBulkFeeResp);
+    if (!estimateFeeResp.suggestedMaxFee || !estimateFeeResp.overall_fee) {
```
stanleyyconsensys, line 123: it looks like a never case , as either it fail in starknet, but it wont return nth

packages/starknet-snap/src/utils/starknetUtils.ts
```diff
@@ -186,6 +186,21 @@ export const declareContract = async (
...
+  const nonceString = await getAccountInstance(
+    network,
+    senderAddress,
+    privateKey,
+    cairoVersion,
+  ).getNonce('latest');
+  return parseInt(nonceString.toLowerCase(), 16);
+};
+
 export const estimateFee = async (
```
stanleyyconsensys, line 204: do we still need this method?

packages/starknet-snap/src/utils/starknetUtils.ts
```diff
@@ -214,13 +234,18 @@ export const estimateFeeBulk = async (
   invocationsDetails?: UniversalDetails,
   cairoVersion?: CairoVersion,
 ): Promise<EstimateFee[]> => {
-  return getAccountInstance(
+  let nonce = invocationsDetails?.nonce;
+  if (!nonce) {
```
stanleyyconsensys, line 238: what happen if the account is new?

stanleyyconsensys (changes requested):
Can you tell me again why we need use latest block on estimate fee , rather than pending? (default)

packages/starknet-snap/src/utils/starknetUtils.ts
```diff
@@ -186,6 +186,21 @@ export const declareContract = async (
   });
 };
 
+export const getAccountNonce = async (
+  network: Network,
+  senderAddress: string,
+  privateKey: string | Uint8Array,
+  cairoVersion?: CairoVersion,
+): Promise<BigNumberish> => {
```
stanleyyconsensys, line 194: we try not to depends on their type too much, 
as your return type is int anyway, it wont become bigint
but im okay to keep BigNumberish in this PR,

packages/starknet-snap/src/utils/starknetUtils.ts
stanleyyconsensys, line 238: !nonce is also true when nonce = 0

stanleyyconsensys (changes requested):

packages/starknet-snap/src/utils/starknetUtils.ts
khanti42: maybe we call it `getAccountLatestNonce` because the method we use says  

`Gets the nonce of the account with respect to a specific block` 

in our case we get `latest`. So we get the nonce of the account with respect to the latest block.

khanti42: @stanleyyconsensys

packages/starknet-snap/src/utils/starknetUtils.ts
khanti42: We keep it for now because it's still used elsewhere.

stanleyyconsensys (changes requested):

packages/starknet-snap/src/utils/starknetUtils.ts
```diff
@@ -214,13 +235,19 @@ export const estimateFeeBulk = async (
   invocationsDetails?: UniversalDetails,
   cairoVersion?: CairoVersion,
 ): Promise<EstimateFee[]> => {
-  return getAccountInstance(
+  // Use nonce from invocationsDetails or get it from the account nonce on the latest accepted block to avoid issues with pending transactions.
+  // as estimating fees on the pending block can fail if there are already transactions in the pending state.
+  const nonce =
+    invocationsDetails?.nonce ??
+    (await getAccountNonceOnLatest(network, senderAddress, privateKey));
+  return await getAccountInstance(
```
stanleyyconsensys, line 243: actually i just find out there is a easier way to do that

by update the getProvider to 
```
export const getProvider = (network: Network, blockIdentifier): ProviderInterface => {
  let providerParam: ProviderOptions = {};
  providerParam = {
    nodeUrl: getRPCUrl(network.chainId),
  };
  if (blockIdentifier) {
providerParam.blockIdentifier = blockIdentifier
}
  return new Provider(providerParam);
};
```
in getAccountInstance
```
export const getAccountInstance = (
  network: Network,
  userAddress: string,
  privateKey: string | Uint8Array,
  cairoVersion?: CairoVersion,
  blockIdentifier?: blockIdentifier
): Account => {
  const provider = getProvider(network, blockIdentifier);
  return new Account(
    provider,
    userAddress,
    privateKey,
    cairoVersion ?? CAIRO_VERSION,
  );
};
```
WDYT

khanti42: That would be great. Can we confirm that the `new Account` uses the blockIdentifier of the provider ?

khanti42: Double checked works like a charm.

stanleyyconsensys: yea, it is actually depends on the provider

so we setup the config of a provider
then we pass provider as an arg to account
when account doing the api, it will use the provider options if no nonce pass in

stanleyyconsensys (approved):
LGTM

PR #290, opened 2024-07-12
fix: get balance should check on pending block

khanti42:
In order for the balance to be in sync we need to read it from pending block instead of latest.


khanti42:
closed - branch naming convention issue. See #293

PR #292, opened 2024-07-16
fix: add additional test cases for toJson function

yuzuswapteam001:
add additional test cases for toJson function


stanleyyconsensys:
we will change it to jest soon

PR #293, opened 2024-07-23
fix: get balance should check on pending block

khanti42:
In order for the balance to be in sync we need to read it from pending block instead of latest.


packages/starknet-snap/src/utils/starknetUtils.ts
```diff
@@ -94,15 +95,19 @@ export const callContract = async (
   contractAddress: string,
   contractFuncName: string,
   contractCallData: RawCalldata = [],
+  blockIdentifier?: BlockIdentifier,
```
stanleyyconsensys, line 98:
```suggestion
  blockIdentifier: BlockIdentifier = 'latest',
```

packages/starknet-snap/src/utils/starknetUtils.ts
```diff
@@ -94,15 +95,19 @@ export const callContract = async (
...
 ): Promise<CallContractResponse> => {
   const provider = getProvider(network);
+  if (!blockIdentifier) {
+    blockIdentifier = 'latest';
+  }
```
stanleyyconsensys, line 103:
```suggestion
```

packages/starknet-snap/src/getErc20TokenBalance.ts
```diff
@@ -35,7 +35,13 @@ export async function getErc20TokenBalance(params: ApiParams) {
 
     logger.log(`getErc20Balance:\nerc20Address: ${erc20Address}\nuserAddress: ${userAddress}`);
 
-    const resp = await callContract(network, erc20Address, 'balanceOf', [num.toBigInt(userAddress).toString(10)]);
+    const resp = await callContract(
+      network,
+      erc20Address,
+      'balanceOf',
+      [num.toBigInt(userAddress).toString(10)],
+      'pending',
```
stanleyyconsensys, line 43: 1. replace above function by `getBalance`
2. i feel like we can have a enum to store 'pending' & 'latest' (base on MM eslint rules, please name the constant to PascalCase)
3. getBalance is always using pending? or sometime latest depends on use case ?

stanleyyconsensys (changes requested):

packages/starknet-snap/src/getErc20TokenBalance.ts
```diff
@@ -50,13 +52,19 @@ export async function getErc20TokenBalance(params: ApiParams) {
       `getErc20Balance:\nerc20Address: ${erc20Address}\nuserAddress: ${userAddress}`,
     );
 
-    const resp = await callContract(network, erc20Address, 'balanceOf', [
-      numUtils.toBigInt(userAddress).toString(10),
-    ]);
+    const blockIdentifier = (await isAccountDeployed(network, userAddress))
+      ? 'pending'
```
stanleyyconsensys, line 56: can we have the pending and latest in constant / enum ? 

e.g 
enum BlockIdentifier {
  Latest: 'latest',
  Pending: 'pending'
}

stanleyyconsensys (changes requested):

packages/starknet-snap/src/getErc20TokenBalance.ts
stanleyyconsensys: btw, why we need isAccountDeployed in here?

stanleyyconsensys: add back unit test

stanleyyconsensys (changes requested):

packages/starknet-snap/src/types/snapState.ts
```diff
@@ -94,4 +94,9 @@ export type Transaction = {
   eventIds: string[];
   timestamp: number;
 };
+
+export enum BlockIdentifierEnum {
```
stanleyyconsensys, line 98: move it to constants.ts 
and For Enum key
we need to use PascalCase

packages/starknet-snap/src/utils/starknetUtils.ts
```diff
@@ -139,6 +140,7 @@ export const callContract = async (
   contractAddress: string,
   contractFuncName: string,
   contractCallData: RawCalldata = [],
+  blockIdentifier: BlockIdentifier = 'latest',
```
stanleyyconsensys, line 143: can we also replace all latest to enum

packages/starknet-snap/test/src/getErc20TokenBalance.test.ts
```diff
@@ -35,9 +35,36 @@ describe('Test function: getErc20TokenBalance', function () {
     sandbox.restore();
   });
 
-  it('should get the ERC-20 token balance correctly', async function () {
+  it('should get the ERC-20 token balance correctly from pending if account is deployed', async function () {
```
stanleyyconsensys, line 38: the test case should be
- should get the ERC-20 token balance correctly
- should get ERC-20 token balance with BlockIdentifier 'pending' if the account is deployed 
- should get ERC-20 token balance with BlockIdentifier 'latest' if the account is not deployed

packages/starknet-snap/src/getErc20TokenBalance.ts
```diff
@@ -50,13 +53,21 @@ export async function getErc20TokenBalance(params: ApiParams) {
       `getErc20Balance:\nerc20Address: ${erc20Address}\nuserAddress: ${userAddress}`,
     );
 
-    const resp = await callContract(network, erc20Address, 'balanceOf', [
-      numUtils.toBigInt(userAddress).toString(10),
-    ]);
+    // Use LATEST block for non-deployed accounts to avoid showing non-zero balances
+    // from pending transactions, as deployment requires a confirmed non-zero balance.
```
stanleyyconsensys, line 57: the comment seem not reference to the use case?
(dont copy, but just the meaning)

case 1: when account is not deployed, we use latest block to make sure the balance is correct/stable, which is the same behaviour when discovering the account is deploy required
case 2: when account is deployed, we use pending block to make sure we can show the balance update asap to unblock concurrent transaction

stanleyyconsensys (changes requested):

stanleyyconsensys (approved):
LGTM

PR #294, opened 2024-07-23
fix: Additional test cases

lotteam002:
add test cases for starknetUtils and snapUtils


stanleyyconsensys:
we will change to Jest soon

PR #295, opened 2024-07-24
refactor: apply MetaMask eslint style into snap

stanleyyconsensys:
This PR is to apply the eslint style from MetaMask to this project to align the same standard


packages/starknet-snap/package.json
```diff
@@ -46,25 +46,39 @@
...
+    "@metamask/snaps-jest": "^8.0.0",
     "@types/chai": "^4.3.1",
     "@types/chai-as-promised": "^7.1.5",
     "@types/sinon": "^10.0.11",
     "@types/sinon-chai": "^3.2.8",
-    "@typescript-eslint/eslint-plugin": "^5.20.0",
-    "@typescript-eslint/parser": "^5.20.0",
+    "@typescript-eslint/eslint-plugin": "^5.42.1",
+    "@typescript-eslint/parser": "^5.42.1",
+    "bip174": "^2.1.1",
```
khanti42, line 61: Do we need bip174 ?

packages/starknet-snap/src/declareContract.ts
```diff
@@ -1,30 +1,48 @@
...
     const requestParamsObj = requestParams as DeclareContractRequestParams;
 
     logger.log(`executeTxn params: ${toJson(requestParamsObj, 2)}}`);
 
-    const senderAddress = requestParamsObj.senderAddress;
+    const { senderAddress } = requestParamsObj;
     const network = getNetworkFromChainId(state, requestParamsObj.chainId);
-    const { privateKey, publicKey } = await getKeysFromAddress(keyDeriver, network, state, senderAddress);
+    const { privateKey, publicKey } = await getKeysFromAddress(
+      keyDeriver,
```
khanti42, line 31: Why we don't do : `keyDeriver as unknown as BIP44AddressKeyDeriver,` here ?

packages/starknet-snap/src/getTransactionStatus.ts
```diff
@@ -1,23 +1,36 @@
...
+    logger.log(
+      `getTransactionStatus:\ngetTxnStatusResp: ${toJson(getTxnStatusResp)}`,
+    );
 
     return getTxnStatusResp;
-  } catch (err) {
-    logger.error(`Problem found: ${err}`);
-    throw err;
+  } catch (error) {
+    // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
```
khanti42, line 32: could we get rid of this disabling by doing something like : 


```
  let errorMessage: string;
  if (error instanceof Error) {
    errorMessage = error.message;
  } else if (typeof error === 'object') {
    errorMessage = JSON.stringify(error);
  } else {
    errorMessage = String(error);
  }
  console.error(`Caught an error: ${errorMessage}`);
  ```

File: packages/starknet-snap/src/index.ts [hunk 4]
```diff
@@ -28,41 +56,37 @@ import {
...
+  getCorrectContractAddress,
+  getKeysFromAddressIndex,
+} from './utils/starknetUtils';
+import { verifySignedMessage } from './verifySignedMessage';
 
 declare const snap;
 const saveMutex = new Mutex();
 
-export const onRpcRequest: OnRpcRequestHandler = async ({ origin, request }) => {
+export const onRpcRequest: OnRpcRequestHandler = async ({ request }) => {
```
khanti42, line 78: Why do we remove `origin` in here ? Is this part of the eslint style ?

packages/starknet-snap/src/index.ts
```diff
@@ -279,34 +308,54 @@ export const onHomePage: OnHomePageHandler = async () => {
...
+    ) {
       network = state.currentNetwork;
     }
 
     // we only support 1 address at this moment
     const idx = 0;
     const keyDeriver = await getAddressKeyDeriver(snap);
-    const { publicKey } = await getKeysFromAddressIndex(keyDeriver, network.chainId, state, idx);
+    const { publicKey } = await getKeysFromAddressIndex(
+      keyDeriver,
```
khanti42, line 322: sometimes we cast it : `keyDeriver as unknown as BIP44AddressKeyDeriver` (for example in : ./estimateAccountDeployFee.ts:35) sometimes we don't. Is there a reason ?

packages/starknet-snap/src/utils/constants.ts
```diff
@@ -33,7 +38,7 @@ export const STARKNET_MAINNET_NETWORK: Network = {
   name: 'Starknet Mainnet',
   chainId: constants.StarknetChainId.SN_MAIN,
   baseUrl: 'https://alpha-mainnet.starknet.io',
-  nodeUrl: 'https://starknet-mainnet.infura.io/v3/60c7253fb48147658095fe0460ac9ee9',
```
khanti42, line 36: Why removing nodeUrl ?

packages/starknet-snap/src/utils/logger.ts
```diff
@@ -9,80 +9,87 @@ export enum LogLevel {
...
 
-  private _logLevel: LogLevel = LogLevel.OFF;
+  readonly #logLevel: LogLevel = LogLevel.OFF;
 
   constructor() {
     this.init(LogLevel.OFF.toString());
   }
 
-  private setLogLevel = function (level: string): void {
+  readonly #setLogLevel = function (level: string): void {
```
khanti42, line 54: do we need readonly for class methods ?

khanti42 (commented):
LGTM, just minor questions / remarks.

packages/starknet-snap/src/index.ts
stanleyyconsensys: i think it is because the keyDeriver pass in from apiparam or get it by urself

packages/starknet-snap/src/utils/constants.ts
stanleyyconsensys: it is no longer use, and try not keep those id in the source code

packages/starknet-snap/src/declareContract.ts
stanleyyconsensys: it was because getKeysFromAddress is using any for keyDeriver, i will change back

packages/starknet-snap/src/index.ts
stanleyyconsensys: yes, unused variable

khanti42 (approved):
LGTM

khanti42 (approved):
LGTM

Issue #277, opened 2024-07-05
Upgrade to starknet.js v 6.11.0


irisdv:
Currently the starknet-snap is using version `6.7.0` of starknet.js. In order to complete #185 we need at least version `6.10.0` as we updated the starknet ID contracts on sepolia and this change was included in `6.10.0`.
Version `6.11.0` is the latest stable version of starknet.js

stanleyyconsensys:
@irisdv we have already upgrade the version to 6.11.0, may i ask if #185 will require any update on snap rpc? if yes , you wanna need to follow the new revamp structure, we will guide you soon

PR #296, opened 2024-07-24
feat: upgrade to sarknet.js v6.11.0

khanti42:
Closes issue #277


stanleyyconsensys (changes requested):
Can we separate the http-server in other PR?
beside what is the reason to have the debug UI? it is quite outdate
and please also update the starknet.js in wallet ui,get-starknet too

packages/starknet-snap/src/utils/keyPair.ts
```diff
@@ -1,7 +1,7 @@
 import type { BIP44AddressKeyDeriver } from '@metamask/key-tree';
 import { getBIP44AddressKeyDeriver } from '@metamask/key-tree';
 import { utils } from 'ethers';
-import { number, ec } from 'starknet_v4.22.0';
+import { num as numUtils, ec } from 'starknet';
```
stanleyyconsensys, line 4: this is very risky change, are we confirm the address is always the same as before

packages/wallet-ui/package.json
```diff
@@ -6,19 +6,19 @@
   "license": "(Apache-2.0 OR MIT)",
   "scripts": {
     "allow-scripts": "yarn workspace root allow-scripts",
-    "build": "react-scripts build",
+    "build": "craco build",
```
stanleyyconsensys, line 9: what is the reason to using craco? when every other dapp in consensys is using react-scripts / gasby

stanleyyconsensys (changes requested):

packages/wallet-ui/package.json
stanleyyconsensys: get-starknet still using 5.14.0

packages/starknet-snap/src/utils/keyPair.ts
khanti42: Ok, because of potential big implication we rollback this and will be done later on an other PR

packages/wallet-ui/package.json
stanleyyconsensys, line 9: Move craco to another PR

stanleyyconsensys (changes requested):

packages/wallet-ui/craco.config.js
```diff
@@ -0,0 +1,20 @@
+module.exports = {
```
stanleyyconsensys, line 1: Craco is not removed

yarn.lock
```diff
@@ -2130,13 +2130,32 @@ __metadata:
...
     sinon-chai: ^3.7.0
-    starknet: 6.7.0
+    starknet: ^6.11.0
     starknet_v4.22.0: "npm:starknet@4.22.0"
     ts-node: ^10.9.2
     typescript: ^4.7.4
   languageName: unknown
   linkType: soft
 
+"@craco/craco@npm:^7.1.0":
```
stanleyyconsensys, line 2140: craco is not removed

stanleyyconsensys (changes requested):

packages/get-starknet/src/snap.ts
```diff
@@ -109,7 +110,7 @@ export class MetaMaskSnap {
     senderAddress: string,
     txnInvocation: AllowArray<Call>,
     abis?: Abi[],
-    invocationsDetails?: InvocationsDetails,
+    invocationsDetails?: UniversalDetails,
   ): Promise<InvokeFunctionResponse> {
```
stanleyyconsensys, line 114: it looks like it is not sync with the snap

stanleyyconsensys (changes requested):

packages/get-starknet/src/snap.ts
stanleyyconsensys: we keep the snap in sync with invocationsDetails for now and fix in the other PR

stanleyyconsensys:
Can we pin the version in this PR?

stanleyyconsensys (dismissed):
LGTM

stanleyyconsensys (approved):

PR #299, opened 2024-07-29
chore: show snap error if it is dev

stanleyyconsensys:
This PR is to show the error message when rpc request fail and log level is not 'OFF'

This PR also change the method for createAccount to have await instead of return, hence to be able to catch the error

Jira: https://consensyssoftware.atlassian.net/jira/software/projects/SF/boards/472?selectedIssue=SF-663

### DIFF
--- a/packages/starknet-snap/src/index.ts
+++ b/packages/starknet-snap/src/index.ts
@@ -6,12 +6,12 @@ import type {
   Component,
 } from '@metamask/snaps-sdk';
 import {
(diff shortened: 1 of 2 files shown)


khanti42 (changes requested):
Could we remove the logging of errors when throwing errors purposely during tests ? 

Capture d’écran 2024-07-29 à 12 13 31

khanti42 (approved):
LGTM

PR #304, opened 2024-07-31
refactor: add account generator for unit test

stanleyyconsensys:
This PR is to add account generator to replace the use of a constant account to produce a more realistic test case

This PR move mock file `snap-provider.mock.ts` to `./packages/starknet-snap/test` , due to it is more convenient to have mock function `manageState` and `dialog` in place. Otherwise the lint will throw error if we try to import an function from a mock file

This PR also adding a demo for `onHomePage` in `index.test.ts` to demo how to use the generate account and the mock state function


khanti42 (approved):
LGTM

PR #307, opened 2024-07-31
refactor: add error utils

stanleyyconsensys:
This PR is to add an error utils to check if the error is an Snap Error

This PR can be combine to use with this  PR https://github.com/Consensys/starknet-snap/pull/305 
When validate request is throwing an Snap Error, and we should not hide the validation error by checking if the error is an snap error or not 

possible use case
```
try {
   validateRequest(params, struct)
} catch (error) {
   if (isSnapRpcError(error) {
       throw error
   }
   throw new SnapError('masked error message')
}
```


khanti42 (approved):
LGTM

PR #308, opened 2024-07-31
chore: return spendable and actual balance in rpc api `getErc20Balance`

khanti42:
## Description
This pull request enhances the `getErc20Balance` function to return both the spendable balance and the actual balance of an ERC-20 token. The primary goal is to provide users with a more accurate representation of their available funds by distinguishing between confirmed (actual) balances and spendable balances, taking into account any pending transactions.

### Function Update
The `getErc20Balance` function now calculates and returns two balances:
- **Spendable Balance:** The minimum of the confirmed balance and the latest state balance (considering pending transactions).
- **Actual Balance:** The confirmed balance without considering pending transactions.

### Balance Calculation
- **Confirmed Balance:** The balance after all confirmed transactions.
- **Latest State Balance:** The balance considering pending transactions (both incoming and outgoing).
- **Spendable Balance:** Calculated as the minimum of the confirmed balance and the latest state balance.

## Motivation
- **User Experience:** By showing the spendable balance, users are provided with a clearer picture of the amount they can actually use, preventing confusion and potential transaction failures due to pending transactions.
- **Accuracy:** Ensures that the displayed balance reflects the true spendable amount, avoiding scenarios where users see tokens that are not yet confirmed and therefore not spendable.

## Testing
- Updated unit tests to cover scenarios with pending transactions.
- Verified that the function correctly returns both the spendable and actual balances.

## Impact
- **Backward Compatibility:** The function signature changes to return two balances instead of one, which may require updates in the consuming code.
- **User Interface:** UIs displaying the token balance will need to be updated to show both the spendable and actual balances.

### Requirement:
- [x] #289


packages/starknet-snap/test/src/getErc20TokenBalance.test.ts
```diff
@@ -51,7 +51,10 @@ describe('Test function: getErc20TokenBalance', function () {
...
     apiParams.requestParams = requestObject;
     const result = await getErc20TokenBalance(apiParams);
-    expect(result).to.be.eq('0x64a');
+    expect(result).to.be.eql({
+      spendableBalance: hexAmount,
+      totalBalance: hexAmount,
+    });
   });
 
   it('should get ERC-20 token balance with BlockIdentifier pending if the account is deployed', async function () {
```
stanleyyconsensys, line 60: test case has to change, as we are not return balance by account is deployed, but the smaller one

packages/wallet-ui/src/services/useStarkNetSnap.ts
```diff
@@ -694,7 +694,7 @@ export const useStarkNetSnap = () => {
           },
         },
       });
-      return response;
+      return response.totalBalance;
```
stanleyyconsensys, line 697: i believe you are trying to do a backward compatible case , so it should be using the latest balance?

packages/starknet-snap/src/getErc20TokenBalance.ts
```diff
@@ -53,23 +49,43 @@ export async function getErc20TokenBalance(params: ApiParams) {
...
+    // - X token balance, initiate a transaction of 0.1X
+    // ==> pending balance 0.9 X, Latest balance X, spendable is 0.9 X
+    // - X token balance, receives 0.1X
+    // ==> pending balance 1.1 X, Latest balance X, spendable is X
+    const balanceBigInt =
+      balancePending < balanceLatest ? balancePending : balanceLatest;
+    const spendableBalance = `0x${balanceBigInt.toString(16)}`;
+    const totalBalance = `0x${balancePending.toString(16)}`;
+    const resp = {
+      spendableBalance,
```
stanleyyconsensys, line 83: i just think, if this api is for our dapp, why not just return the balancePending and balanceLatest
and let the FE to determined what balance they need to show?

as backend (snap), it may be better to give 2 informations and let the caller to choose their preference?

stanleyyconsensys (changes requested):

packages/starknet-snap/src/getErc20TokenBalance.ts
stanleyyconsensys: and btw , i think it is better to return non hex value ? (this api was not created by me, but i think most of the api is returning an string number "123123213123124124" rather than "0x123123")
or your wanna do it in other PR?

khanti42: do you mean to return the integer version not the hex version, e.g. instead of '0xa' we return '10' ?

packages/starknet-snap/src/getErc20TokenBalance.ts
```diff
@@ -53,23 +49,26 @@ export async function getErc20TokenBalance(params: ApiParams) {
...
+      userAddress,
+      erc20Address,
+      network,
+      BlockIdentifierEnum.Pending,
     );
 
-    logger.log(`getErc20Balance:\nresp: ${toJson(balance)}`);
+    const resp = {
+      balancePending,
+      balanceLatest,
```
stanleyyconsensys, line 67: how about change the response to MM standard? we may need to discuss

API Response reference - https://github.com/MetaMask/keyring-api/blob/main/src/api/keyring.ts

Caip 19 asset id - https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-19.md

e.g 
# DAI Token
starknet:SN_GOERLI/erc20:0x6b175474e89094c44da98b954eedeac495271d0f

packages/starknet-snap/src/getErc20TokenBalance.ts
stanleyyconsensys: yea, it seem more readable

stanleyyconsensys: but it is up to us at this moment

stanleyyconsensys (approved):

packages/starknet-snap/src/getErc20TokenBalance.ts
khanti42: I agree, let's match MM standard. But let's do this in separate PR if ok with you.

PR #310, opened 2024-07-31
chore: display both spendable and actual balance in wallet-ui

khanti42:
### Description
This update modifies the user interface to display both the spendable balance and the actual balance of ERC-20 tokens. The goal is to provide users with a clear distinction between the balance they can immediately use and their total confirmed balance, considering any pending transactions.

### Requirement:
- [x] #308


packages/wallet-ui/src/services/useStarkNetSnap.ts
```diff
@@ -694,7 +722,11 @@ export const useStarkNetSnap = () => {
           },
         },
       });
-      return response;
+      if (balanceType === BalanceType.Spendable) {
+        return response.spendableBalance;
+      } else {
+        return response.totalBalance;
+      }
```
stanleyyconsensys, line 729: IMO, the method should just return both response, else you may need to query twice to get both value, which is not make sense

packages/wallet-ui/src/services/useStarkNetSnap.ts
```diff
@@ -663,10 +683,17 @@ export const useStarkNetSnap = () => {
         accountAddress,
         chainId,
       );
+      const tokenBalanceSpendable = await getTokenBalance(
+        tokenAddress,
+        accountAddress,
+        chainId,
+        BalanceType.Spendable,
+      );
```
stanleyyconsensys, line 691: this is not necessary

packages/wallet-ui/src/services/useStarkNetSnap.ts
```diff
@@ -624,10 +637,17 @@ export const useStarkNetSnap = () => {
           accountAddress,
           chainId,
         );
+        const tokenBalanceSpendable = await getTokenBalance(
+          tokenAddress,
+          accountAddress,
+          chainId,
+          BalanceType.Spendable,
+        );
```
stanleyyconsensys, line 645: this is not necessary

packages/wallet-ui/src/services/useStarkNetSnap.ts
```diff
@@ -277,6 +277,18 @@ export const useStarkNetSnap = () => {
...
+        const accountAddr = Array.isArray(acc) ? acc[0].address : acc.address;
+        return await getTokenBalance(
+          token.address,
+          accountAddr,
+          chainId,
+          BalanceType.Spendable,
+        );
+      }),
+    );
+
```
stanleyyconsensys, line 291: this is not necessary

packages/wallet-ui/src/utils/utils.ts
```diff
@@ -45,22 +45,31 @@ export const isValidAddress = (toCheck: string) => {
...
   balance?: string,
+  balanceSpendable?: string,
   usdPrice?: number,
 ): Erc20TokenBalance => {
   return {
     ...token,
     amount: ethers.BigNumber.from(balance ? balance : '0x0'),
+    spendableAmount: ethers.BigNumber.from(
+      balanceSpendable ? balanceSpendable : '0x0',
+    ),
```
stanleyyconsensys, line 56: this is not necessary

stanleyyconsensys (changes requested):
In the UI part, we dont need to another getSpendableTotalBalance , as it will create twice api call
the api already return 2 values

packages/wallet-ui/src/utils/utils.ts
```diff
@@ -218,3 +248,39 @@ export const shortenDomain = (domain: string, maxLength = 18) => {
   const shortenedPartLength = maxLength - ellipsis.length;
   return `${domain.substring(0, shortenedPartLength)}${ellipsis}`;
 };
+
+export function getTokenBalanceWithDetails(
+  tokenBalance: TokenBalance,
+  token: Erc20Token,
+  tokenUSDPrice: any,
```
stanleyyconsensys, line 255: according to `addMissingPropertiesToToken` i think it is number | undefined

```suggestion
  tokenUSDPrice?: number,
```

stanleyyconsensys (changes requested):
some code improvement advice

packages/wallet-ui/src/utils/utils.ts
```diff
@@ -218,3 +248,39 @@ export const shortenDomain = (domain: string, maxLength = 18) => {
...
+): Erc20TokenBalance {
+  const { balancePending, balanceLatest } = tokenBalance;
+  const tokenTotalBalance =
```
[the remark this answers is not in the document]
khanti42: agreed. Note that If `tokenTotalBalance = balanceLatest`, we lose the detailed view of the total amount including incoming pending transactions, but I agree it reduces user confusion.

packages/wallet-ui/src/utils/utils.ts
```diff
@@ -218,3 +247,19 @@ export const shortenDomain = (domain: string, maxLength = 18) => {
...
   return `${domain.substring(0, shortenedPartLength)}${ellipsis}`;
 };
+
+export function getTokenBalanceWithDetails(
+  tokenBalance: TokenBalance,
+  token: Erc20Token,
+  tokenUSDPrice?: number,
+): Erc20TokenBalance {
+  const { balancePending, balanceLatest } = tokenBalance;
+  const spendableBalance =
```
stanleyyconsensys, line 257: nit: 

```suggestion
  const spendableBalance = Math.min(balancePending, balanceLatest)
```

packages/wallet-ui/src/utils/utils.ts
```diff
@@ -45,22 +50,31 @@ export const isValidAddress = (toCheck: string) => {
...
 
 export const getHumanReadableAmount = (
   asset: Erc20TokenBalance,
   assetAmount?: string,
+  balanceType?: BalanceType,
 ) => {
+  const amount =
+    balanceType === BalanceType.Spendable
+      ? asset.spendableAmount ?? asset.amount
+      : asset.amount;
```
stanleyyconsensys, line 74: i think this method has been used in some place, and mainly purpose is to format the given value, 
assetAmount is a var to override the assert val if needed

e.g when a tx history display, it call this method and override the token val (balance) by the tx val

so i think may be not good to add a logic here to see if it is spendable, 

do you think it is better to put those logic from caller?

packages/wallet-ui/src/utils/utils.ts
```diff
@@ -79,6 +93,21 @@ export const getHumanReadableAmount = (
...
+    asset,
+    assetAmount,
+    BalanceType.Spendable,
+  );
+  const totalAmount = getHumanReadableAmount(asset);
+  return spendableAmount === totalAmount
+    ? `${spendableAmount}`
+    : `${spendableAmount} (${totalAmount})`;
+};
+
```
stanleyyconsensys, line 110: as`getSpendableTotalBalance` is only used for a case to display the balance, hence the assetAmount doesnt really needed

```suggestion
export const getSpendableTotalBalance = (
  asset: Erc20TokenBalance
): string => {
  const spendableAmount = getHumanReadableAmount(
    asset,
    asset.spendableAmount
  );

  if (asset.spendableAmount === asset.total) {
      return `${spendableAmount}`
  }
  
  const totalAmount = getHumanReadableAmount(
    asset,
    asset.amount
  );
  
   return `${spendableAmount} (${totalAmount})`;
};

```

stanleyyconsensys (changes requested):
LGTM, but i forgot to leave some comment before, so if you dont mind, plz take a look again

packages/wallet-ui/src/utils/utils.ts
khanti42: Math.min does not work with BigNumber

packages/wallet-ui/src/utils/utils.ts
khanti42: agreed, moved up.

packages/wallet-ui/src/utils/utils.ts
```diff
@@ -79,6 +83,19 @@ export const getHumanReadableAmount = (
...
+  const spendableAmount = getHumanReadableAmount(asset, amount.toString());
+
+  if (asset.spendableAmount === asset.amount) {
+    return `${spendableAmount}`;
+  }
+
+  const totalAmount = getHumanReadableAmount(asset, asset.amount.toString());
+
+  return `${spendableAmount} (${totalAmount})`;
+};
```
stanleyyconsensys, line 97: const amount = asset.spendableAmount ?? asset.amount;
it may create a weird case 

if the asset does not have spendableAmount, so it use asset.amount

and `if (asset.spendableAmount === asset.amount) ` will be false

then at the end , it will become `return `${spendableAmount} (${totalAmount})`;`

stanleyyconsensys: i will say do something like that 
```suggestion
  if (asset.spendableAmount === undefined) {
       throw new Error("Spendable amount can not be undefined")
  }
  
  const spendableAmount = getHumanReadableAmount(asset, asset.spendableAmount);

  if (asset.spendableAmount === asset.amount) {
    return `${spendableAmount}`;
  }

  const totalAmount = getHumanReadableAmount(asset, asset.amount.toString());

  return `${spendableAmount} (${totalAmount})`;
};
```

stanleyyconsensys (changes requested):
LGTM, 1 last comment on `getSpendableTotalBalance `

stanleyyconsensys (approved):
LGTM

PR #311, opened 2024-08-01
refactor: add `verifyIfAccountNeedUpgradeOrDeploy` method to group the account deploy or upgrade required validation logic

stanleyyconsensys:
This PR is to add new method `verifyIfAccountNeedUpgradeOrDeploy`  in snapUtils to group the logic of
- show update/deploy alert dialog (optional)
- check if the account needed to upgrade or deploy
- throw upgrade or deploy required error

it also update the `showDeployRequestModal ` and `showUpgradeRequestModal ` to use the alertDialog utils from `snap`

it also remove some of the test case that depends on the removed method `showAccountRequireUpgradeOrDeployModal`, 
as the test responsibility is on the util itself, rather than the rpc api

### DIFF
--- a/packages/starknet-snap/test/src/declareContract.test.ts
+++ b/packages/starknet-snap/test/src/declareContract.test.ts
@@ -18,7 +18,6 @@ import {
 import { getAddressKeyDeriver } from '../../src/utils/keyPair';
 import { Mutex } from 'async-mutex';
 import {
-  ApiParams,
   ApiParamsWithKeyDeriver,
   DeclareContractRequestParams,
 } from '../../src/types/snapApi';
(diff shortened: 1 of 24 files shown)


packages/starknet-snap/test/src/declareContract.test.ts
khanti42, line 1: We are reducing the test coverage on mocha, does that require to reduce the coverage limit in package.json to not break ci/cd ?

khanti42 (changes requested):
Just wondering regarding test cases that have been removed by the PR and their effect. Should we reduce the required test coverage on mocha in `starknet-snap/package.json` as some test are moved to jest ?

packages/starknet-snap/test/src/declareContract.test.ts
stanleyyconsensys: good point, lets reduce the coverage

stanleyyconsensys: i think it does not reduce many coverage, even i remove all the require deploy/upgrade checking

as in previous code, deploy/upgrade checking logic has to try catch on each file , so ur test code needed do the test on every file

but right now, i remove the try catch and move that into a share function `verifyIfAccountNeedUpgradeOrDeploy `
Therefore, in each file, this function will execute anyway, hence it has already replace the coverage

however, for the individual function `verifyIfAccountNeedUpgradeOrDeploy`  , we didnt do the test on mocha side but just the jest side, indeed it will reduce a bit on the coverage, but should be very very minor

stanleyyconsensys: PR: https://github.com/Consensys/starknet-snap/pull/313

khanti42 (approved):

PR #315, opened 2024-08-13
chore: add state management base class and shared lock

stanleyyconsensys:
This PR is to add state management base class `SnapStateManager`

it provides the functionally of `update, get, set, withTransaction, commit, rollback`

- `get` and `set` is the simple method to get the state and set the state.

- `update` is an method combine with get, then set, but also compatible with transaction mode.

- `withTransaction` is a method to init a transaction by getting the state, and do the callback, if anything happen, it will rollback to the state that we get previously, it can combine with`update` to use.

- `commit` is a method to force commit in a transaction before it auto commit.

- `rollback` is a internal method to rollback the state.

example use:
```typescript
export type SnapState = {
  walletIds: string[];
  wallets: Wallets;
};

class MyState extends SnapStateManager<SnapState> {
      constructor() {
        super(true);
      }

     async updateWallet(wallet) {
         await this.update(async (state: SnapState) => {
                 const { id, address } = wallet.account;
                 if (
                    this.isAccountExist(state, id) ||
                    this.getAccountByAddress(state, address)
                  ) {
                    throw new Error(`Account address ${address} already exists`);
                  }
              state.wallets[id] = wallet;
        })
     }
}


// Test rollback
async function testRollback() {
    const mgt = new MyState()
    await mgt.withTransaction(async () => {
        await mgt.addWallet({
          account: keyringAccount,
        });
        // if this line fail, it will not update the state
        await this.#emitEvent(KeyringEvent.AccountCreated, {
          account: keyringAccount,
          accountNameSuggestion: this.getKeyringAccountNameSuggestion(options),
        });
    });
}
```

### DIFF
--- a/packages/starknet-snap/package.json
+++ b/packages/starknet-snap/package.json
@@ -45,7 +45,8 @@
     "ethers": "^5.5.1",
     "starknet": "6.11.0",
     "starknet_v4.22.0": "npm:starknet@4.22.0",
-    "superstruct": "^2.0.2"
+    "superstruct": "^2.0.2",
+    "uuid": "^10.0.0"
   },
   "devDependencies": {
     "@babel/preset-typescript": "^7.23.3",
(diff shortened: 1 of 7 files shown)


khanti42 (approved):
LGTM!

PR #320, opened 2024-08-14
chore: add sharable rpc handler 

stanleyyconsensys:
This PR is to add a sharable rpc handler to have some share workflow across different rpc method.

`RpcController` is an abstract class provide the basic need of an RPC handler, 
- when extends this class, it will apply the workflow of request & response validation by using the internal method `preExecute` and `postExecute`. 
- Which also provide the extensibility to override those workflow if needed

`AccountRpcController` is another abstract class to expends `RpcController`
- it apply the workflow of getting account from the address


khanti42 (approved):
LGTM!

PR #323, opened 2024-08-16
refactor: add `AccountStateManager`

stanleyyconsensys:
This PR is to add `AccountStateManager` to manage the account data in state

This PR also create an Filter Class to allow us to have data filter in OOP

usage:
```typescript
class MyFilter
  extends Filter<string, AccAcontract> {
  protected _apply(data: AccAcontract): boolean {
    return (
       data.address === this.search
    );
  }
}

const mgr = new AccountStateManager()
const result = await mgr.list([
  new MyFilter("address")
])

```


khanti42 (approved):
LGTM!

PR #324, opened 2024-08-19
refactor: add `NetworkStateManager`

stanleyyconsensys:
This PR is to add `NetworkStateManager` as similar as this https://github.com/Consensys/starknet-snap/pull/323

this manager allow caller to:
- update, list, get the network
- set, get the current network
- add default networks by given a list of network to check if it has to update / insert

this PR also adding a test helper  in './state/__tests__/helper.ts' to reduce the duplication of the test code across all state manager

### DIFF
--- a/packages/starknet-snap/src/state/network-state-manager.test.ts
+++ b/packages/starknet-snap/src/state/network-state-manager.test.ts
@@ -11,7 +11,7 @@ import { StateManagerError } from './state-manager';
 
 describe('NetworkStateManager', () => {
   describe('findNetwork', () => {
-    it('returns the token', async () => {
+    it('returns the network', async () => {
       const chainId = constants.StarknetChainId.SN_SEPOLIA;
       await mockState({
         networks: [STARKNET_MAINNET_NETWORK, STARKNET_SEPOLIA_TESTNET_NETWORK],
@@ -25,7 +25,7 @@ describe('NetworkStateManager', () => {
       expect(result).toStrictEqual(STARKNET_SEPOLIA_TESTNET_NETWORK);
     });
 
-    it('returns null if the token can not be found', async () => {
+    it('returns null if the network can not be found', async () => {
       const chainId = constants.StarknetChainId.SN_SEPOLIA;
       await mockState({
         networks: [STARKNET_MAINNET_NETWORK],
@@ -112,7 +112,7 @@ describe('NetworkStateManager', () => {
       expect(result).toStrictEqual([STARKNET_SEPOLIA_TESTNET_NETWORK]);
     });
 
-    it('returns empty array if the account address can not be found', async () => {
+    it('returns empty array if a network with the given chainId cannot be found', async () => {
       const chainId = constants.StarknetChainId.SN_SEPOLIA;
       await mockState({
         networks: [STARKNET_MAINNET_NETWORK],
--- a/packages/starknet-snap/src/state/network-state-manager.test.ts
+++ b/packages/starknet-snap/src/state/network-state-manager.test.ts
@@ -1,0 +1,243 @@
+import { constants } from 'starknet';
+
+import type { Network } from '../types/snapState';
+import {
+  STARKNET_MAINNET_NETWORK,
+  STARKNET_SEPOLIA_TESTNET_NETWORK,
+} from '../utils/constants';
+import { mockState } from './__tests__/helper';
+import { NetworkStateManager, ChainIdFilter } from './network-state-manager';
+import { StateManagerError } from './state-manager';
+
+describe('NetworkStateManager', () => {
+  describe('findNetwork', () => {
+    it('returns the token', async () => {
+      const chainId = constants.StarknetChainId.SN_SEPOLIA;
+      await mockState({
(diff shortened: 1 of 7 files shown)


packages/starknet-snap/src/state/network-state-manager.test.ts
```diff
@@ -0,0 +1,243 @@
...
+  STARKNET_MAINNET_NETWORK,
+  STARKNET_SEPOLIA_TESTNET_NETWORK,
+} from '../utils/constants';
+import { mockState } from './__tests__/helper';
+import { NetworkStateManager, ChainIdFilter } from './network-state-manager';
+import { StateManagerError } from './state-manager';
+
+describe('NetworkStateManager', () => {
+  describe('findNetwork', () => {
+    it('returns the token', async () => {
```
khanti42, line 14: replace "returns the token" by "returns the chainId" ?

packages/starknet-snap/src/state/network-state-manager.test.ts
```diff
@@ -0,0 +1,243 @@
...
+
+      const stateManager = new NetworkStateManager();
+      const result = await stateManager.findNetwork({
+        chainId,
+      });
+
+      expect(result).toStrictEqual(STARKNET_SEPOLIA_TESTNET_NETWORK);
+    });
+
+    it('returns null if the token can not be found', async () => {
```
khanti42, line 28: replace "returns the token" by "returns the chainId" ?

packages/starknet-snap/src/state/network-state-manager.test.ts
```diff
@@ -0,0 +1,243 @@
...
+
+  describe('list', () => {
+    it('returns the list of network by chainId', async () => {
+      const chainId = constants.StarknetChainId.SN_SEPOLIA;
+      await mockState({
+        networks: [STARKNET_MAINNET_NETWORK, STARKNET_SEPOLIA_TESTNET_NETWORK],
+      });
+
+      const stateManager = new NetworkStateManager();
+      const result = await stateManager.list([new ChainIdFilter([chainId])]);
```
khanti42, line 110: can there really be several network with same id ?

packages/starknet-snap/src/state/network-state-manager.test.ts
```diff
@@ -0,0 +1,243 @@
...
+
+      expect(result).toStrictEqual([STARKNET_SEPOLIA_TESTNET_NETWORK]);
+    });
+
+    it('returns empty array if the account address can not be found', async () => {
```
khanti42, line 115: account address => network chainId. 

`it('returns empty array if a network with the given chainId cannot be found')`

packages/starknet-snap/src/state/network-state-manager.test.ts
```diff
@@ -0,0 +1,243 @@
...
+        currentNetwork: STARKNET_MAINNET_NETWORK,
+        updateTo: STARKNET_MAINNET_NETWORK,
+      },
+      {
+        status: 'different',
+        currentNetwork: STARKNET_MAINNET_NETWORK,
+        updateTo: STARKNET_SEPOLIA_TESTNET_NETWORK,
+      },
+    ])(
+      'set the current network when the current network is $status',
```
khanti42, line 206: what is the use case of `status` variable, I don't see it being used ?

packages/starknet-snap/src/state/network-state-manager.ts
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  protected updateEntity(dataInState: Network, data: Network): void {
+    assert(dataInState.name, string());
+
+    dataInState.name = data.name;
+    dataInState.nodeUrl = data.nodeUrl;
+  }
+
+  async addDefaultNetworks(networks: Network[]): Promise<void> {
```
khanti42, line 26: The name could be improved because it can do both update or add depending on if the network exists.

khanti42 (changes requested):

packages/starknet-snap/src/state/network-state-manager.test.ts
stanleyyconsensys: returns the network

packages/starknet-snap/src/state/network-state-manager.test.ts
stanleyyconsensys: no, but the list function is a method from parent class allow you to search by different condition, chainId is 1 of those

as you can see the state has 1 network
i can search the state from 1 chainId or 2 chainIds, and return 1 or more result

packages/starknet-snap/src/state/network-state-manager.test.ts
stanleyyconsensys: it is a generic var replace to this test, if you run the test, it will become "when the current network is same/different/undefined"

packages/starknet-snap/src/state/network-state-manager.ts
stanleyyconsensys: it is not anupsert, but just add a set of default networks, thats why i add this name

let me add a document

khanti42 (approved):

PR #325, opened 2024-08-20
refactor: add `TokenStateManager` 

stanleyyconsensys:
This PR is to add `TokenStateManager` to manage the token state as similar as this https://github.com/Consensys/starknet-snap/pull/323

this manager allow caller to:
- update, list, get the token
- get eth token by chain id
- get strj token by chain id
- add default tokens by given a list of tokens to check if it has to update / insert


khanti42 (approved):
LGTM

PR #328, opened 2024-08-21
refactor: add `TransactionStateManager`

stanleyyconsensys:
This PR is to add TransactionStateManager to manage the txn state as similar as this https://github.com/Consensys/starknet-snap/pull/323

this manager allow caller to:

- update, add , get the txn
- list the txns by different search conditions, by `[Txn Type, Execution Status, Finality Status, Timestamp, Contract Address, Send Address, Chain Id, Txn Hash ]`
- remove the txns by the same search conditions as above

### DIFF
--- a/packages/starknet-snap/src/state/transaction-state-manager.ts
+++ b/packages/starknet-snap/src/state/transaction-state-manager.ts
@@ -1,0 +1,305 @@
+import type { constants, TransactionType } from 'starknet';
+import {
+  TransactionFinalityStatus,
+  TransactionExecutionStatus,
+} from 'starknet';
+import { assert, enums, number } from 'superstruct';
+
+import type { Transaction, SnapState } from '../types/snapState';
+import { TransactionStatusType } from '../types/snapState';
+import type { IFilter } from './filter';
+import {
+  BigIntFilter,
+  ChainIdFilter as BaseChainIdFilter,
+  StringFllter,
+  Filter,
+} from './filter';
+import { StateManager, StateManagerError } from './state-manager';
+
+export type ITxFilter = IFilter<Transaction>;
+
+export class ChainIdFilter
+  extends BaseChainIdFilter<Transaction>
+  implements ITxFilter {}
+
+export class ContractAddressFilter
+  extends BigIntFilter<Transaction>
+  implements ITxFilter
+{
+  dataKey = 'contractAddress';
+}
+export class SenderAddressFilter
+  extends BigIntFilter<Transaction>
+  implements ITxFilter
+{
+  dataKey = 'senderAddress';
+}
+
+export class TxHashFilter
+  extends BigIntFilter<Transaction>
+  implements ITxFilter
+{
+  dataKey = 'txnHash';
+}
+
+export class TxTimestampFilter
+  extends Filter<number, Transaction>
+  implements ITxFilter
+{
+  _apply(data: Transaction): boolean {
+    // The timestamp from the data source is in seconds, but we are comparing it in milliseconds
+    return this.search !== undefined && data.timestamp * 1000 >= this.search;
+  }
+}
+
+export class TxnTypeFilter
+  extends StringFllter<Transaction>
+  implements ITxFilter
+{
+  dataKey = 'txnType';
+}
+
+export class TxStatusFilter implements ITxFilter {
+  finalityStatus: Set<string>;
+
+  executionStatus: Set<string>;
+
+  constructor(finalityStatus: string[], executionStatus: string[]) {
+    this.finalityStatus = new Set(
+      finalityStatus.map((val) => val.toLowerCase()),
+    );
+    this.executionStatus = new Set(
+      executionStatus.map((val) => val.toLowerCase()),
+    );
+  }
+
+  apply(txn: Transaction): boolean {
+    let finalityStatusCond = false;
+    let executionStatusCond = false;
+
+    if (this.finalityStatus.size > 0) {
+      finalityStatusCond =
+        Object.prototype.hasOwnProperty.call(
+          txn,
+          TransactionStatusType.FINALITY,
+        ) &&
+        txn[TransactionStatusType.FINALITY] &&
+        this.finalityStatus.has(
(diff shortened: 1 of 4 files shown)


packages/starknet-snap/src/state/transaction-state-manager.ts
```diff
@@ -0,0 +1,305 @@
...
+  extends BigIntFilter<Transaction>
+  implements ITxFilter
+{
+  dataKey = 'contractAddress';
+}
+export class SenderAddressFilter
+  extends BigIntFilter<Transaction>
+  implements ITxFilter
+{
+  dataKey = 'senderAddress';
```
khanti42, line 35: shall we call senderAddress simply address ? As we don't always send tokens, sometimes we execute for instance.

packages/starknet-snap/src/state/transaction-state-manager.ts
```diff
@@ -0,0 +1,305 @@
...
+  dataKey = 'txnHash';
+}
+
+export class TxTimestampFilter
+  extends Filter<number, Transaction>
+  implements ITxFilter
+{
+  _apply(data: Transaction): boolean {
+    // The timestamp from the data source is in seconds, but we are comparing it in milliseconds
+    return this.search !== undefined && data.timestamp * 1000 >= this.search;
```
khanti42, line 51: here we look for all transaction that happened after the search timestamp ? E.g. if I search for 1000 seconds in the past I put 1000 in timestamp when calling findTransactions ?

packages/starknet-snap/src/state/transaction-state-manager.ts
```diff
@@ -0,0 +1,305 @@
...
+}
+
+export class TxnTypeFilter
+  extends StringFllter<Transaction>
+  implements ITxFilter
+{
+  dataKey = 'txnType';
+}
+
+export class TxStatusFilter implements ITxFilter {
```
khanti42, line 62: Quite complex filter might be good to add some comments

packages/starknet-snap/src/state/transaction-state-manager.ts
```diff
@@ -0,0 +1,305 @@
...
+  /**
+   * Finds a transaction object based on the given chain id and txn hash.
+   *
+   * @param param - The param object.
+   * @param param.txnHash - The txn hash of the transaction object to search for.
+   * @param [param.chainId] - The optional chain id of the transaction object to search for.
+   * @param [state] - The optional SnapState object.
+   * @returns A Promise that resolves with the transaction object if found, or null if not found.
+   */
+  async findTransaction(
```
khanti42, line 208: maybe we call this getTransaction, as it is very different from findTransactions above. Having similar name could be confusing.

packages/starknet-snap/src/state/transaction-state-manager.ts
```diff
@@ -0,0 +1,305 @@
...
+    }
+  }
+
+  /**
+   * Removes the transaction objects in the state with the given search conditions.
+   *
+   * @param searchFilter - The searchFilter object.
+   * @returns A Promise that resolves when the remove is complete.
+   */
+  async removeTransactions(searchFilter: SearchFilter): Promise<void> {
```
khanti42, line 288: What is use case to removeTransactions ?

packages/starknet-snap/test/utils.ts
```diff
@@ -121,3 +125,135 @@ export async function generateAccounts(
...
+ *
+ * @param params
+ * @param params.chainId - Starknet Chain Id.
+ * @param params.address - Address of the account.
+ * @param params.contractAddresses - Contract addresses to generate transactions.
+ * @param params.txnTypes - Array of transaction types.
+ * @param params.finalityStatuses - Array of transaction finality status.
+ * @param params.executionStatuses - Array of transaction execution status.
+ * @param params.cnt - Number of transaction to generate.
+ * @returns An array of StarknetAccount object.
```
khanti42, line 140: Do we really return an array of StarknetAccount object ?? Or an array of transactions ?

packages/starknet-snap/test/utils.ts
```diff
@@ -121,3 +125,135 @@ export async function generateAccounts(
...
+ */
+export function generateTransactions({
```
khanti42, line 142: Why generateTransactions does not use addTransaction ?

khanti42 (changes requested):

packages/starknet-snap/test/utils.ts
stanleyyconsensys: yes it should be transactions

packages/starknet-snap/test/utils.ts
stanleyyconsensys: what does it mean?

packages/starknet-snap/src/state/transaction-state-manager.ts
stanleyyconsensys: remove outdate txn in state, please refer to getTransations, at the end we remove txn from state, as we only display a limit of time of the txns

packages/starknet-snap/src/state/transaction-state-manager.ts
stanleyyconsensys: yup, so lets change all in other state manager too

packages/starknet-snap/src/state/transaction-state-manager.ts
stanleyyconsensys: the filter allow you to search the txn that larger than or equal to the given timestamp

e.g i wanna list txn from today till to 7 days ago

packages/starknet-snap/src/state/transaction-state-manager.ts
stanleyyconsensys: thats include to change the existing txn state from the snap, as well as the type

i would suggest do it in other PR, as this PR is to have mgr to fetch the data from current state 

but i will add a remark on our snapState type

khanti42 (approved):

PR #331, opened 2024-08-23
chore: maintain the naming convention for get single entity in all state manager

stanleyyconsensys:
This PR is to fix the comments on

> maybe we call this getTransaction, as it is very different from findTransactions above. Having similar name could be confusing.
> yup, so lets change all in other state manager too in other PR

_Originally posted by @khanti42 in https://github.com/Consensys/starknet-snap/pull/328#discussion_r1724820935_

and

move the test helper from
`./test/utils.ts`
to
`./src/__tests__/helper.ts`

and

add new **testing util method** `generateRandomValue` to generate a security-sensitive random value that recommended by sonarcloud.

### DIFF
--- a/packages/starknet-snap/src/__tests__/helper.ts
+++ b/packages/starknet-snap/src/__tests__/helper.ts
@@ -1,7 +1,11 @@
+import {
+  BIP44CoinTypeNode,
+  getBIP44AddressKeyDeriver,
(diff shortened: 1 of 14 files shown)


khanti42 (approved):
LGTM, maybe let's just add in description of PR the small changes regarding random number generation to have a record.

PR #332, opened 2024-08-23
refactor: revamp `starkNet_verifySignedMessage`

stanleyyconsensys:
This PR is to revamp `starkNet_verifySignedMessage` to adopt the sharable rpc handler from this PR https://github.com/Consensys/starknet-snap/pull/320

This PR also change the input of the `signatureStr`, which to align with the output from rpc `starkNet_signMessage`
from:
```
name :  `signatureStr` 
type: `string`
```
to:
```
name :  `signature` 
type: `Array<Hex>`
```
This PR also include `HexStruct` from `@metamask/utils`


khanti42 (approved):
LGTM

PR #333, opened 2024-08-23
refactor: revamp `starkNet_signDeclareTransaction`

stanleyyconsensys:
This PR is to revamp `starkNet_signDeclareTransaction` to adopt the sharable rpc handler from this PR https://github.com/Consensys/starknet-snap/pull/320


khanti42 (approved):
LGTM

packages/starknet-snap/src/rpcs/sign-declare-transaction.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  /**
+   * Execute the sign declare transaction request handler.
+   * It will show a confirmation dialog to the user before signing the declare transaction.
+   *
+   * @param params - The parameters of the request.
+   * @param params.address - The address of the signer.
+   * @param params.details - The declare transaction details to sign.
+   * @param [params.enableAuthorize] - Optional, a flag to enable or display the confirmation dialog to the user.
+   * @param params.chainId - The chain id of the network.
```
khanti42, line 59: Why do we have chainId in here ? Isn't this already in this.network.chainId

stanleyyconsensys: this has to set up as the parent class is using the chainId, please refer to the parent class to know more

khanti42 (approved):
LGTM

PR #334, opened 2024-08-24
feat: Addition of `createStructWithAdditionalProperties` for Incremental Validation of Complex StarkNet Types

khanti42:
This PR introduces the `createStructWithAdditionalProperties` utility function that allows to combine a predefined set of properties with dynamically added properties, enabling more granular control over type validation without requiring full validation upfront. Its designed to facilitate the validation of complex StarkNet types in a flexible and iterative manner.

Example :

```
const predefined = object({
  name: string(),
  age: number(),
});

const MyStruct = createStructWithAdditionalProperties(predefined);

const data = {
  name: 'John',
  age: 30,
  nickname: 'Johnny', // This will be validated as `any` since `additionalPropertyTypes` defaults to `any`.
};

const [error, result] = MyStruct.validate(data);
if (error) {
  console.error(error);
} else {
  console.log(result); // { name: 'John', age: 30, nickname: 'Johnny' }
}
```

### DIFF
--- a/packages/starknet-snap/src/utils/superstruct.test.ts
+++ b/packages/starknet-snap/src/utils/superstruct.test.ts
@@ -1,5 +1,5 @@
 import { constants } from 'starknet';
-import { StructError, assert } from 'superstruct';
+import { StructError, assert, object, number, string } from 'superstruct';
 
 import transactionExample from '../__tests__/fixture/transactionExample.json';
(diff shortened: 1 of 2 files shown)


stanleyyconsensys (changes requested):
what is the use case for addtional properties

as we should not allow extra property to be added?

stanleyyconsensys (approved):

PR #335, opened 2024-08-24
fix: improve error message in rpc input validation

khanti42:
#### Before:
When testing a `union` struct with multiple levels, such as:

```typescript
const multiLevelStruct = union([
  object({
    field1: string(),
  }),
  object({
    field3: boolean(),
  }),
]);
```

If the input was incorrect, for example:

```json
{
  "field1": true
}
```

The error message provided was not clear and didn't offer much guidance:

```
"Expected the value to satisfy a union of `object | object`, but received: [object Object]"
```

This generic message made it difficult to diagnose the specific issue within the nested structure, particularly when working with complex schemas.

#### Now:
The error handling has been improved to provide much clearer and more informative messages. With the same incorrect input:

```json
{
  "field1": true
}
```

The new error message provides detailed information about where the validation failed:

```
At path: field1 --
    Expected a string, but received: true
    Expected a value of type `never`, but received: `true`

At path: field3 -- 
    Expected a value of type `boolean`, but received: `undefined`
```

#### Benefits:
- **Enhanced Clarity**: The error messages now pinpoint the exact field and expected type, making it easier to debug and correct invalid input.
- **Detailed Feedback**: By providing detailed feedback at each level of the validation, developers can quickly identify and resolve issues with their input data.
- **Better Support for Complex Schemas**: This improvement is especially valuable when dealing with nested or complex data structures, where understanding the root cause of validation failures is crucial.

This  PR improves  the developer experience and reduce time in debugging.

### DIFF
--- a/packages/starknet-snap/src/utils/rpc.test.ts
+++ b/packages/starknet-snap/src/utils/rpc.test.ts
@@ -1,6 +1,6 @@
 import { InvalidParamsError, SnapError } from '@metamask/snaps-sdk';
 import { constants } from 'starknet';
-import { object, string } from 'superstruct';
+import { boolean, object, string, union } from 'superstruct';
 import type { Struct, Infer } from 'superstruct';
 
 import type { StarknetAccount } from '../__tests__/helper';
@@ -30,6 +30,15 @@ const validateParam = {
     '0x04882a372da3dfe1c53170ad75893832469bf87b62b13e84662565c4a88f25cd',
 };
 
+const treeStruct = union([
+  object({
+    field1: string(),
+  }),
+  object({
+    field3: boolean(),
+  }),
+]);
+
 describe('validateRequest', () => {
   it('does not throw error if the request is valid', () => {
     expect(() =>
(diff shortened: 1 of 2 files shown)


packages/starknet-snap/src/utils/rpc.test.ts
```diff
@@ -30,6 +30,15 @@ const validateParam = {
     '0x04882a372da3dfe1c53170ad75893832469bf87b62b13e84662565c4a88f25cd',
 };
 
+const treeStruct = union([
```
stanleyyconsensys, line 33: it is better to use superstruct refine, rather than on RPC or
use the union from metamask/utils

stanleyyconsensys (changes requested):

khanti42:
This is closed, because complex superstruct are implemented as : https://github.com/MetaMask/keyring-api/blob/c7068d1d585c116cd8cd2aa2e838429215800de5/src/api.ts#L63-L76 instead of unions.

PR #339, opened 2024-08-29
chore: refine execute txn

stanleyyconsensys:
This PR is to refine the execute txn rpc


packages/wallet-ui/src/components/ui/organism/Header/SendSummaryModal/SendSummaryModal.view.tsx
```diff
@@ -95,7 +95,7 @@ export const SendSummaryModalView = ({
           wallet.accounts[0] as unknown as string,
           chainId,
           selectedFeeToken === FeeToken.ETH
-            ? constants.TRANSACTION_VERSION.V2
+            ? constants.TRANSACTION_VERSION.V1
             : constants.TRANSACTION_VERSION.V3,
```
stanleyyconsensys, line 99: let make everywhere:, if we pay strk, we assign version3, else, do not assign any version

PR #340, opened 2024-08-29
fix: transaction version v2 does not apply

khanti42:
This PR is to change the management of transaction version. 

- The account instance will not set any transaction version. 
- Transaction version are set by the caller function. 
- Default transaction is 0x1.
- Available transaction version is 0x1, 0x3 (0x2 should not be used)

### DIFF
--- a/packages/starknet-snap/src/rpcs/executeTxn.ts
+++ b/packages/starknet-snap/src/rpcs/executeTxn.ts
@@ -117,6 +117,7 @@ export class ExecuteTxnRpc extends AccountRpcController<
         address,
         accountDeployed,
         calls,
+        suggestedMaxFee,
         details?.version as unknown as constants.TRANSACTION_VERSION,
       ))
     ) {
--- a/packages/starknet-snap/src/rpcs/executeTxn.ts
+++ b/packages/starknet-snap/src/rpcs/executeTxn.ts
@@ -168,19 +168,20 @@ export class ExecuteTxnRpc extends AccountRpcController<
     transactionHash: string,
   ): Promise<void> {
     if (!transactionHash) {
-      throw new Error('Failed to deploy account');
+      throw new Error(`Failed to deploy account for address ${address}`);
     }
-
-    await this.txnStateManager.withTransaction(async (_) => {
-      await this.txnStateManager.addTransaction(
-        this.createDeployTxn(address, transactionHash),
-      );
-      await this.accStateManager.updateAccountAsDeploy({
-        address,
-        chainId: this.network.chainId,
-        transactionHash,
-      });
-    });
+    // TODO uncomment this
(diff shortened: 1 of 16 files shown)


packages/starknet-snap/src/rpcs/executeTxn.ts
```diff
@@ -117,6 +117,7 @@ export class ExecuteTxnRpc extends AccountRpcController<
         address,
         accountDeployed,
         calls,
+        suggestedMaxFee,
```
khanti42, line 120: this a small miss I found, not related to v1/v2 change but the UX only

packages/starknet-snap/src/utils/starknetUtils.ts
```diff
@@ -1079,7 +1075,7 @@ export async function getEstimatedFees(
     suggestedMaxFee: estimateFeeResp.suggestedMaxFee.toString(10),
     overallFee: estimateFeeResp.overall_fee.toString(10),
     unit:
-      transactionVersion === constants.TRANSACTION_VERSION.V2
+      invocationsDetails.version === constants.TRANSACTION_VERSION.V1
```
stanleyyconsensys, line 1078:
```suggestion
      invocationsDetails?.version === constants.TRANSACTION_VERSION.V3 ? FeeTokenUnit.STRK: FeeTokenUnit.ETH
```

it is better to if it is v3, then strk, everything else eth

packages/starknet-snap/src/utils/starknetUtils.ts
```diff
@@ -1052,8 +1049,8 @@ export async function getEstimatedFees(
   transactionInvocations: Invocations,
   invocationsDetails: UniversalDetails = {},
 ): Promise<EstimateFeeResponse> {
-  const transactionVersion = (invocationsDetails?.version ??
-    TRANSACTION_VERSION) as TransactionVersion;
+  invocationsDetails.version =
+    invocationsDetails?.version ?? TRANSACTION_VERSION;
```
stanleyyconsensys, line 1053:
```suggestion
```

packages/starknet-snap/src/utils/superstruct.ts
```diff
@@ -109,11 +109,11 @@ export const NumberStringStruct = union([number(), HexStruct]);
 export const CairoVersionStruct = enums([CAIRO_VERSION, CAIRO_VERSION_LEGACY]);
 
 export const TxVersionStruct = enums([
-  constants.TRANSACTION_VERSION.V2,
+  constants.TRANSACTION_VERSION.V1,
```
stanleyyconsensys, line 112: on signDeclareTransaction, it only accept v2 / v3 ^^

stanleyyconsensys (changes requested):

PR #341, opened 2024-09-02
feat: add `StarkScan` client to fetch transaction history

stanleyyconsensys:
This PR is adding a client sdk to fetch the transaction history with StarkScan

This is the first journey to enable fetching transaction by StarkScan

This PR includes:
- a starkscan data client to enable:
   - getTransactions (by address , by till to timestamp)
   - getDeployTransaction (by address)
- a dataClient interface `IDataClient` for generic interface
- a factory to create the `StarkScanClient`
- a config mapping for `StarkScanClient`


packages/starknet-snap/src/types/snapState.ts
```diff
@@ -134,4 +143,26 @@ export type Transaction = {
...
+  contractAddress: string; // in hex
+  executionStatus?: TransactionExecutionStatus | string;
+  finalityStatus?: TransactionFinalityStatus | string;
+  failureReason: string;
+  timestamp: number;
+  maxFee?: string | null;
+  actualFee?: string | null;
+  // using Record<string, TranscationAccountCall[]> to support O(1) searching
+  accountCalls?: Record<string, TranscationAccountCall[]> | null;
+  version: 'V2';
```
khanti42, line 160: What about V3 transactions ? 

https://starkscan.readme.io/reference/transaction-object

packages/starknet-snap/snap.config.ts
```diff
@@ -15,6 +15,7 @@ const config: SnapConfig = {
     SNAP_ENV: process.env.SNAP_ENV ?? 'prod',
     VOYAGER_API_KEY: process.env.VOYAGER_API_KEY ?? '',
```
khanti42, line 16: When will we discontinue VOYAGER ?

packages/starknet-snap/src/chain/data-client/starkscan.ts
```diff
@@ -0,0 +1,280 @@
...
+      senderAddress: this.getSenderAddress(tx),
+      timestamp,
+      finalityStatus,
+      executionStatus,
+      maxFee,
+      actualFee,
+      contractAddress: this.getContractAddress(tx),
+      accountCalls,
+      failureReason: failureReason ?? '',
+      version: 'V2',
```
khanti42, line 234: Why don't we include version from the response here ?

khanti42 (changes requested):
Some clarifying comments

packages/starknet-snap/src/chain/data-client/starkscan.ts
```diff
@@ -0,0 +1,280 @@
...
+
+  /**
+   * Fetches the deploy transaction for a given contract address.
+   *
+   * @param address - The address of the contract to fetch the deploy transaction for.
+   * @returns A Promise that resolve the Transaction object or null if the transaction can not be found.
+   */
+  async getDeployTransaction(address: string): Promise<Transaction | null> {
+    // Fetch the first 5 transactions to find the deploy transaction
+    // The deploy transaction usually is the first transaction from the list
```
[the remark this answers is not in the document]
stanleyyconsensys: it is for buffer, some very very legacy account, may use declare before deploy

packages/starknet-snap/src/types/snapState.ts
stanleyyconsensys: it is not for TX v3, it is for separate the type of the state data

packages/starknet-snap/snap.config.ts
stanleyyconsensys: until the get-stransactions fully released , then we are safe to remove

packages/starknet-snap/src/chain/data-client/starkscan.ts
stanleyyconsensys: yea you are right, may be it is better to include a version of the transaction

khanti42 (approved):
LGTM

PR #342, opened 2024-09-03
refactor: revamp logger and set logger level from snap build

stanleyyconsensys:
This PR is to refactor the logger utils

it includes:
- revamp logger to using number [0-6] rather than string to indicate the log level to reduce the conflict of using string as log level
- add unit test for logger
- log level setup by snap build, but no longer pass from client, as we are expecting log message should not render from production, using the previous handle may cause security concern 
- add temp solution (global hook) for mocha unit test to by pass the logger mock, we can remove this after we are fully migrate to jest
- add `Config` constant to group the snap configuration (today, we only have loglevel, but we will migrate others later)


packages/starknet-snap/src/utils/logger.test.ts
```diff
@@ -0,0 +1,78 @@
+import { logger, LogLevel } from './logger';
+
+describe('Logger', () => {
```
khanti42, line 3: here I guess we only test the log level that we actually use, this is why the following are not tested ? 

```
  ERROR = 1,
  WARN = 2,
  DEBUG = 4,
  TRACE = 5,
 ```

File: packages/starknet-snap/src/utils/logger.ts [hunk 2]
```diff
@@ -32,35 +32,31 @@ export const emptyLog: LoggingFn = (
   // eslint-disable-next-line @typescript-eslint/no-empty-function
   {};
 
-class Logger implements ILogger {
-  readonly log: LoggingFn;
+export class Logger implements ILogger {
+  log: LoggingFn;
 
-  readonly warn: LoggingFn;
+  warn: LoggingFn;
```
khanti42, line 38: What is the reason of removing the readonly ? 
- Is this to be able to change log on the fly if needed ? 
- Or since now log is set by the snap env file and not from ui this is not needed anymore ?

packages/starknet-snap/test/src/index.test.ts
```diff
@@ -87,36 +84,6 @@ describe('onRpcRequest', function () {
       );
     }
   });
-
-  it('does not hide the error message if an error is thrown and `LogLevel` is not `OFF`', async function () {
```
khanti42, line 91: I guess this is handled in a new way in next PR ?

packages/starknet-snap/src/utils/logger.test.ts
stanleyyconsensys: lets add it back

packages/starknet-snap/src/utils/logger.ts
```diff
@@ -32,35 +32,31 @@ export const emptyLog: LoggingFn = (
...
-  readonly trace: LoggingFn;
+  trace: LoggingFn;
 
-  readonly #logLevel: LogLevel = LogLevel.OFF;
+  #logLevel: LogLevel = LogLevel.OFF;
 
-  constructor() {
-    this.init(LogLevel.OFF.toString());
+  set logLevel(level: LogLevel) {
+    this.#logLevel = level;
```
khanti42, line 51: what happens if we set a logLevel that is not valid ? Maybe we could use a superstruct ?

packages/starknet-snap/src/utils/logger.ts
stanleyyconsensys: prev code , readonly is incorrect, as we are updating in init()

packages/starknet-snap/test/src/index.test.ts
stanleyyconsensys: yes

packages/starknet-snap/src/utils/logger.ts
stanleyyconsensys: log level is not from client side, no need

khanti42 (approved):
LGTM

PR #343, opened 2024-09-03
refactor: revamp snap `index.ts` error handle

stanleyyconsensys:
This PR is to refactor the behaviour on how `index.ts` handle the error

The actual error message should not mask if it is expected to throw , e.g SnapError
Else it should mask with an general error message
Rather than depends on the log level to show / hide the actual error

The PR changes include:

- change error handle
  - from:
    - throw general error message if log level === off
    - throw actual error message if log level !== off
  - to:
    - throw general error message if the error is not a Snap Error
    - throw error message if the error is a Snap Error (including validation error)

- remove duplicate error handle layer from `RpcController`

### Requirement:

- [x] #342


packages/starknet-snap/src/index.test.ts
```diff
@@ -50,15 +51,30 @@ describe('onRpcRequest', () => {
...
+    const { createAccountSpy } = createMockSpy();
+    createAccountSpy.mockRejectedValue(new SnapError('error'));
 
-    await expect(onRpcRequest(createMockRequest())).rejects.toThrow(
-      'Unable to execute the rpc request',
-    );
+    await expect(onRpcRequest(createMockRequest())).rejects.toThrow(SnapError);
+  });
+
+  it('throws `SnapError` if the error is not an instance of SnapError', async () => {
```
khanti42, line 73: This wording is a bit misleading. Does this mean we always throw SnapError ?

khanti42: Maybe we should link it with the log level ?

packages/starknet-snap/src/index.ts
```diff
@@ -293,19 +295,18 @@ export const onRpcRequest: OnRpcRequestHandler = async ({ request }) => {
...
-      ) as unknown as Error;
-    } else {
-      throw new SnapError(error.message) as unknown as Error;
+    let snapError = error;
+
+    if (!isSnapRpcError(error)) {
+      snapError = new SnapError('Unable to execute the rpc request');
     }
+    logger.error(
+      `onRpcRequest error: ${JSON.stringify(snapError.toJSON(), null, 2)}`,
```
khanti42, line 307: Shall we add a test for this logger.error to ensure that it's shown when log level is error ?

packages/starknet-snap/src/index.test.ts
stanleyyconsensys: yes we always throw SnapError , it has nothing to do with log level

packages/starknet-snap/src/index.ts
stanleyyconsensys: no need, as the logger test is already tested the feature

khanti42 (approved):
LGTM

PR #344, opened 2024-09-04
refactor: revamp `snapUtils` function - `dappUrl` 

stanleyyconsensys:
This PR is to change `snapUtils` function - `dappUrl` to using config value rather than env directly, to reduce the noice of process.env

This PR includes:
- move dappUrl to new utils file - `url.ts` and rename `dappUrl` to `getDappUrl`
- using enum rather than string to indicate  snap env, when it is not possible to pass in from client
- add snap env into `Config`.`snapEnv`  with default value


packages/starknet-snap/src/utils/url.ts
khanti42, line 1: Do we expect to have other url managed in url.ts ?
If it's only for the dapp why not calling it `dappUrl`, wdyt ?

khanti42 (commented):
Shall we add an example in the env sample file ?

packages/starknet-snap/src/utils/url.ts
stanleyyconsensys: i think this is an helper for url, may be there will be some more url utils to come in, and it is very often case

stanleyyconsensys: what do u mean by example, if u are talking about "SNAP_ENV",  doesnt it already included in the env

khanti42 (dismissed):
LGTM

khanti42 (approved):
LGTM

PR #350, opened 2024-09-09
feat: set up snap for jsx support 

khanti42:
## Summary

This PR introduces support for JSX in the `starknet-snap` project, following the official [MetaMask Snaps JSX upgrade tutorial](https://docs.metamask.io/snaps/features/custom-ui/with-jsx/#upgrade-an-existing-snap-to-use-jsx). The changes include updates to ESLint configuration, package linting, and type dependencies to ensure proper handling of `.tsx` files across the project.

## Changes

1. **ESLint Configuration (`.eslintrc.js`):**
   - Adjusted the `@typescript-eslint/no-shadow` rule to allow shadowing of `Text`.

2. **Package Scripts (`package.json`):**
   - Updated the `lint:eslint` command to include `.tsx` files.
   - Updated the `lint:misc` command to format `.tsx` files as well.

3. **Dependency Updates (`yarn.lock`):**
   - Added support for React types and JSX, including:
     - Upgrading `@types/react` from `18.3.3` to `18.0.15`.
     - Adding `@types/scheduler` for proper scheduling and type handling in React components.
   - Updated the lock file to ensure compatibility with the new changes.

4. **Renaming of `index.ts` into `index.tsx` and `index.test.ts` and `index.test.tsx`** 
   - In order to prepare for next PR #351 we rename `index.ts` to `index.tsx` and update `packages/starknet-snap/snap.config.ts`

## Additional Notes

- These changes are essential to supporting JSX in the `starknet-snap` package, making it easier to use JSX components within Snaps. Following this PR, JSX syntax can now be leveraged for improved UI development within the Snap environment.
- Please ensure that the `.tsx` support works seamlessly across the project before merging.

PR #351, opened 2024-09-09
feat: updates onHomePage onInstall and onUpdate to use jsx component 

khanti42:
## Description 

Replaces old style components with jsx based components in 'index.tsx' and updates the tests accordingly in 'index.test.tsx'

## Requirements 

- [ ] #350


package.json
```diff
@@ -25,6 +25,7 @@
     "@commitlint/config-conventional": "^17.0.3",
     "@lavamoat/allow-scripts": "^3.0.0",
     "@lavamoat/preinstall-always-fail": "^2.0.0",
+    "@types/react": "18.0.15",
```
khanti42, line 28: this is to solve the vscode issue of getting false positive types from react 18.0.15

khanti42:
Will re-open later.

PR #352, opened 2024-09-10
fix: QA review request, show only one balance

khanti42:
**PR Description: Simplifying Balance Display for Better User Clarity**

**Context:**
This PR addresses feedback from QA regarding the dual balance display, which was considered potentially confusing for users. The current display shows two balances: the spendable balance and the total available funds (e.g., 100 (102)). The feedback highlighted that displaying both values could cause users to misinterpret their actual spendable balance, especially in scenarios where the spendable balance is 0, but there are pending funds. 

**Changes Implemented:**
- The dual balance display has been simplified by showing only the balance on the pending block, ensuring the balance is updated after each transaction, this is needed because of the ability to submit multiple transaction in the same block.


packages/wallet-ui/src/utils/utils.ts
```diff
@@ -87,20 +83,15 @@ export const getHumanReadableAmount = (
 };
 
 export const getSpendableTotalBalance = (asset: Erc20TokenBalance): string => {
```
stanleyyconsensys, line 85: nit: i guess this method can be remove and rollback to getHumanReadableAmount from the caller? but im also okay to keep this here

stanleyyconsensys (approved):
LGTM

PR #354, opened 2024-09-17
chore: ensure backward compatibility with remoteEntry.js of snap v2.9.0

khanti42:
### PR Description:

**Backward Compatibility for Request Parameters**

This PR ensures backward compatibility with version 2.9.0 for StarkNet-related methods (`starknet_sign*`, `starknet_execute`, and `starknet_declare`). 

- **`signerAddress` to `address` mapping**: In older versions, `starknet_sign*` methods used the `signerAddress` field. This update ensures that if `signerAddress` is passed in the request, it is mapped to the `address` field required by the new version.
  
- **`senderAddress` to `address` mapping**: Similarly, for `starknet_execute` and `starknet_declare`, which previously used `senderAddress`, the PR adds logic to map this to `address`.

By handling both `signerAddress` and `senderAddress`, this update allows for smooth transitions to the new parameter structure without breaking existing integrations.

### DIFF
--- a/packages/starknet-snap/src/index.ts
+++ b/packages/starknet-snap/src/index.ts
@@ -63,7 +63,7 @@ import type {
 } from './types/snapApi';
 import type { SnapState } from './types/snapState';
 import { upgradeAccContract } from './upgradeAccContract';
-import { getDappUrl, isSnapRpcError, mapDeprecatedParams } from './utils';
+import { getDappUrl, isSnapRpcError } from './utils';
 import {
   CAIRO_VERSION_LEGACY,
   ETHER_MAINNET,
@@ -152,21 +152,6 @@ export const onRpcRequest: OnRpcRequestHandler = async ({ request }) => {
       wallet: snap,
       saveMutex,
     };
-
-    // Define mappings to ensure backward compatibility with previous versions of the API.
-    // These mappings replace deprecated parameter names with the updated equivalents,
-    // allowing older integrations to function without changes, while the new version
-    // expects parameters such as `calls`, `details`, and `address`.
-    const paramMappings: Record<string, string> = {
-      signerAddress: 'address',
-      senderAddress: 'address',
-      txnInvocation: 'calls',
-      invocationsDetails: 'details',
-      transaction: 'details',
-    };
-
-    // Apply the mappings to apiParams.requestParams
-    mapDeprecatedParams(apiParams.requestParams, paramMappings);
 
     switch (request.method) {
       case 'starkNet_createAccount':
--- a/packages/starknet-snap/src/index.ts
+++ b/packages/starknet-snap/src/index.ts
@@ -63,7 +63,7 @@ import type {
 } from './types/snapApi';
 import type { SnapState } from './types/snapState';
 import { upgradeAccContract } from './upgradeAccContract';
-import { getDappUrl, isSnapRpcError } from './utils';
+import { getDappUrl, isSnapRpcError, mapDeprecatedParams } from './utils';
 import {
   CAIRO_VERSION_LEGACY,
   ETHER_MAINNET,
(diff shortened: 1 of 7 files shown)


packages/starknet-snap/src/index.ts
```diff
@@ -153,6 +153,21 @@ export const onRpcRequest: OnRpcRequestHandler = async ({ request }) => {
...
+    const paramMappings: Record<string, string> = {
+      signerAddress: 'address',
+      senderAddress: 'address',
+      txnInvocation: 'calls',
+      invocationsDetails: 'details',
+      transaction: 'details',
+    };
+
+    // Apply the mappings to apiParams.requestParams
+    mapDeprecatedParams(apiParams.requestParams, paramMappings);
```
stanleyyconsensys, line 169: the mapping is override every api, as some api is using the same params, as the new API is using OOP, why not implement in those API?

packages/starknet-snap/src/utils/formatterUtils.ts
```diff
@@ -7,3 +7,33 @@ export const hexToString = (hexStr) => {
...
+ *   signerAddress: 'address',
+ *   senderAddress: 'address',
+ *   txnInvocation: 'calls',
+ *   invocationsDetails: 'details',
+ *   transaction: 'details'
+ * };
+ * mapDeprecatedParams(apiParams.requestParams, paramMappings);
+ */
+export const mapDeprecatedParams = (
+  requestParams: any,
```
stanleyyconsensys, line 29:
```suggestion
  requestParams: Record<string, Json>,
```

packages/starknet-snap/src/utils/formatterUtils.ts
```diff
@@ -7,3 +7,33 @@ export const hexToString = (hexStr) => {
...
+  mappings: Record<string, string>,
+) => {
+  Object.keys(mappings).forEach((oldParam) => {
+    const newParam = mappings[oldParam];
+    if (requestParams[oldParam]) {
```
stanleyyconsensys, line 34: nit:
```suggestion
    if (Object.prototype.hasOwnProperty.call(requestParams, oldParam)) {
```

stanleyyconsensys (commented):
unit test is not added in the utils

packages/starknet-snap/src/utils/formatterUtils.ts
```diff
@@ -7,3 +9,33 @@ export const hexToString = (hexStr) => {
...
+  mappings: Record<string, string>,
+) => {
+  Object.keys(mappings).forEach((oldParam) => {
+    const newParam = mappings[oldParam];
+    if (Object.prototype.hasOwnProperty.call(requestParams, oldParam)) {
+      requestParams[newParam] = requestParams[oldParam];
+      delete requestParams[oldParam]; // Remove old param after mapping
+    }
+  });
+};
```
stanleyyconsensys, line 41:
```suggestion
export const mapDeprecatedParams = <Params>(
  requestParams: Params
  mappings: Record<string, string>,
) => {
  Object.keys(mappings).forEach((oldParam) => {
    const newParam = mappings[oldParam]  as unknown as keyof Params;
    if (Object.prototype.hasOwnProperty.call(requestParams, oldParam)) {
      requestParams[newParam] = requestParams[oldParam];
      delete requestParams[oldParam]; // Remove old param after mapping
    }
  });
};
```

nit:
in that case, we dont need to cast the type in execute txn

packages/starknet-snap/src/utils/formatterUtils.test.ts
```diff
@@ -0,0 +1,67 @@
...
+      txnInvocation: 'calls',
+    };
+
+    mapDeprecatedParams(requestParams, mappings);
+
+    expect(requestParams).not.toHaveProperty('signerAddress');
+    expect(requestParams).not.toHaveProperty('txnInvocation');
+  });
+
+  it('should not map if deprecated parameter does not exist', () => {
```
stanleyyconsensys, line 40:
```suggestion
  it('does nothing if the deprecated parameter does not exist', () => {
```

stanleyyconsensys (changes requested):
LGTM, just some minor comment on the test

stanleyyconsensys (approved):
LGTM

PR #357, opened 2024-09-20
feat: default network should be main net

khanti42:
This PR changes the default network from Sepolia to Mainnet and updates the corresponding tests. 

It also resets the default to mainnet on install.

### DIFF
--- a/packages/starknet-snap/src/index.ts
+++ b/packages/starknet-snap/src/index.ts
@@ -55,6 +55,7 @@ import {
 } from './rpcs';
 import { sendTransaction } from './sendTransaction';
 import { signDeployAccountTransaction } from './signDeployAccountTransaction';
+import { NetworkStateManager } from './state/network-state-manager';
 import { switchNetwork } from './switchNetwork';
 import type {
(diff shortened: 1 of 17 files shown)


packages/starknet-snap/src/index.ts
```diff
@@ -354,11 +354,10 @@ export const onHomePage: OnHomePageHandler = async () => {
     }
 
     // default network is testnet
-    let network = STARKNET_SEPOLIA_TESTNET_NETWORK;
-
+    let network = STARKNET_MAINNET_NETWORK;
     if (
       state.currentNetwork &&
-      state.currentNetwork.chainId !== STARKNET_TESTNET_NETWORK.chainId
+      state.currentNetwork.chainId !== STARKNET_MAINNET_NETWORK.chainId
```
stanleyyconsensys, line 360: this line is for detect if the user using a deprecated network, so we dont need to change it

stanleyyconsensys (changes requested):

stanleyyconsensys (approved):
LGTM

PR #360, opened 2024-09-30
chore: update feature change notice dialog for 2.10.0

khanti42:


packages/starknet-snap/src/index.ts
```diff
@@ -328,7 +328,14 @@ export const onInstall: OnInstallHandler = async () => {
...
     text('Features released with this update:'),
-    text('Cairo contract upgrade support.'),
+    row(
+      'Gas Fee',
+      text(
+        'Support STRK token for the gas fee in sending transaction and estimating fee.',
+      ),
+    ),
+    row('Network', text('Default network changed to mainnet.')),
+    row('Transactions', text('Support for multiple consecutive transactions.')),
```
stanleyyconsensys, line 338: i would suggest no need to add left label, but just this text?

stanleyyconsensys (changes requested):

stanleyyconsensys (approved):

PR #366, opened 2024-10-04
fix(get-starknet): address not update when network change

stanleyyconsensys:
this PR is to fix the address not update when network change

it is due to an user may have 2 different address when using different network due to different cairo contract version / contract upgrade


khanti42 (approved):
LGTM!

PR #369, opened 2024-10-08
refactor(NetworkStateManager): update `NetworkStateManager` with new configuration

stanleyyconsensys:
This PR is to update the `getNetwork` , `updateNetwork` & `getCurrentNetwork` logic in `NetworkStateManager`

it changes:
 `getNetwork` -> look up configuration -> available network 
`updateNetwork` -> as  `getNetwork` update with fallback look up, the update method has to update the state, so it search the entity by using the underlying `find` instead
`getCurrentNetwork` -> return configuration default network

### DIFF
--- a/packages/starknet-snap/src/state/network-state-manager.ts
+++ b/packages/starknet-snap/src/state/network-state-manager.ts
@@ -61,7 +61,7 @@ export class NetworkStateManager extends StateManager<Network> {
   /**
    * Finds a network based on the given chainId.
    * The query will first be looked up in the state. If the result is false, it will then fallback to the available Networks constants.
-   * 
+   *
    * (Note) Due to the returned network object may not exist in the state, it may failed to execute `updateNetwork` with the returned network object.
    *
    * @param param - The param object.
--- a/packages/starknet-snap/src/state/network-state-manager.ts
+++ b/packages/starknet-snap/src/state/network-state-manager.ts
@@ -1,10 +1,10 @@
 import { assert, string } from 'superstruct';
 
+import { Config } from '../config';
 import type { Network, SnapState } from '../types/snapState';
 import type { IFilter } from './filter';
 import { ChainIdFilter as BaseChainIdFilter } from './filter';
 import { StateManager, StateManagerError } from './state-manager';
-import { Config } from '../config';
 
 export type INetworkFilter = IFilter<Network>;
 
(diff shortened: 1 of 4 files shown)


packages/starknet-snap/src/state/network-state-manager.ts
```diff
@@ -88,10 +94,9 @@ export class NetworkStateManager extends StateManager<Network> {
   async updateNetwork(data: Network): Promise<void> {
```
khanti42, line 94: Is it intentional that a network can be retrieved from `availableNetworks` but not updated if it’s not in the state? If so, why is this behavior preferable?
If the goal is to ensure that any network retrieved can also be updated, should the state be updated with the network when it’s first accessed in `getNetwork` (if it’s not found)?
Otherwise we could comment this behavior, so future maintainers know that not all networks can be updated unless they exist in the state?

packages/starknet-snap/src/utils/snapUtils.ts
```diff
@@ -34,7 +35,6 @@ import {
   MAXIMUM_TOKEN_SYMBOL_LENGTH,
   PRELOADED_NETWORKS,
   PRELOADED_TOKENS,
-  STARKNET_MAINNET_NETWORK,
   STARKNET_SEPOLIA_TESTNET_NETWORK,
```
khanti42, line 38: Shall we also remove this:  
 
```suggestion
```

And then use `constants.StarknetChainId.SN_SEPOLIA` instead of `STARKNET_SEPOLIA_TESTNET_NETWORK.chainId` for consistency ?

khanti42 (changes requested):

packages/starknet-snap/src/utils/snapUtils.ts
stanleyyconsensys: not should what you referring to, but i believe it is not related to this PR? 
but indeed, i agree if we know the network, we dont need to use our own constant to get the ID

packages/starknet-snap/src/state/network-state-manager.ts
stanleyyconsensys: or we can remove update method, as we dont have that use case

khanti42: I agree, let's remove this. If we need it we add it back. It then removes the confusion.

khanti42 (dismissed):
LGTM

packages/starknet-snap/src/state/network-state-manager.ts
stanleyyconsensys: just find out we have a use case 
we cant remove it actually

so update the comment on getNetwork instead

PR #370, opened 2024-10-08
feat(get-starknet): add wallet_supportedSpecs rpc method

khanti42:
## Requirements:

- [x] : #365


packages/get-starknet/src/constants.ts
```diff
@@ -1,5 +1,8 @@
...
+
+export const SupportedStarknetSpecVersion = ['0.7'];
```
stanleyyconsensys, line 8:
```suggestion
// The supported RPC version should be update base on the Provider.
// With Provider `Alchemy`, it is 0.7
export const SupportedStarknetSpecVersion = ['0.7'];
```

stanleyyconsensys (changes requested):
Test is missing

PR #371, opened 2024-10-08
chore: get-starknet linter relax

khanti42:
Exclude JavaScript files from ESLint checks to resolve lint errors on webpack configuration files.

Requirements: 

- [x] : #365


packages/get-starknet/.eslintrc.js
```diff
@@ -38,5 +38,5 @@ module.exports = {
     },
   ],
 
-  ignorePatterns: ['!.eslintrc.js', 'dist/', '**/test', '.nyc_output/', 'coverage/'],
+  ignorePatterns: ['!.eslintrc.js', 'dist/', '**/test', '.nyc_output/', 'coverage/', '*.js'],
```
stanleyyconsensys, line 41: may be just for webpack?

as in root directory, we already has a rule to check js file

```suggestion
  ignorePatterns: ['!.eslintrc.js', 'dist/', '**/test', '.nyc_output/', 'coverage/', 'webpack.*.js'],
```

stanleyyconsensys (changes requested):

PR #372, opened 2024-10-08
feat(get-starknet): add wallet_supportedWalletApi rpc method 

khanti42:


packages/get-starknet/src/rpcs/supported-wallet-api.ts
```diff
@@ -0,0 +1,14 @@
...
+import { SupportedWalletApi } from '../constants';
+import { StarknetWalletRpc } from '../utils/rpc';
+
+export type WalletSupportedWalletApiMethod = 'wallet_supportedWalletApi';
+type Params = RpcTypeToMessageMap[WalletSupportedWalletApiMethod]['params'];
+type Result = RpcTypeToMessageMap[WalletSupportedWalletApiMethod]['result'];
+
+export class WalletSupportedWalletApi extends StarknetWalletRpc {
+  async handleRequest(_param: Params): Promise<Result> {
+    return SupportedWalletApi;
```
stanleyyconsensys, line 12: i think the response format is not correct?

it is
``` 
   [{
      api_version: "some version"
   }]
```

packages/get-starknet/src/constants.ts
```diff
@@ -1,10 +1,16 @@
...
 
 // The supported RPC version should be update base on the Provider.
 // With Provider `Alchemy`, it is 0.7
 export const SupportedStarknetSpecVersion = ['0.7'];
+
+// The wallet API support is 0.7.2 but the RPC specs requests xx.yy. Hence we skip the last digits.
```
stanleyyconsensys, line 15: i believe WalletApi is different with RPC
it is talking about the wrapper API version (wallet api versions)

packages/get-starknet/src/type.ts
```diff
@@ -30,3 +30,8 @@ export type RequestSnapResponse = {
     blocked: boolean;
   };
 };
+
+export type ApiVersion = {
```
stanleyyconsensys, line 34: it seem useless?

packages/get-starknet/src/constants.ts
```diff
@@ -1,10 +1,16 @@
+import type { ApiVersion } from './type';
```
stanleyyconsensys, line 1: starknet.js contains that type

stanleyyconsensys (changes requested):
Test is missing

packages/get-starknet/src/rpcs/supported-wallet-api.ts
khanti42: Based on what I read here : https://github.com/starkware-libs/starknet-specs/blob/6d88b7399f56260ece3821c71f9ce53ec55f830b/wallet-api/wallet_rpc.json#L19 It seems to be an array of strings not array of object.

stanleyyconsensys: we can use string[] for now , but await for clarify

packages/get-starknet/src/constants.ts
stanleyyconsensys: we can use [0.7] for now , but await for clarify

packages/get-starknet/src/rpcs/supported-wallet-api.test.ts
```diff
@@ -0,0 +1,15 @@
...
+
+describe('WalletSupportedWalletApi', () => {
+  it('returns the supported wallet api version', async () => {
+    const wallet = createWallet();
+    mockWalletInit({ currentNetwork: SepoliaNetwork });
+
+    const walletSupportedWalletApi = new WalletSupportedWalletApi(wallet);
+    const result = await walletSupportedWalletApi.execute();
+
+    expect(result).toBe(SupportedWalletApi);
```
stanleyyconsensys, line 13:
```suggestion
    expect(result).toStrictEqual(SupportedWalletApi);
```

stanleyyconsensys (commented):
a minor comment in the test

stanleyyconsensys (approved):

PR #375, opened 2024-10-10
refactor(get-starknet): add get-starknet error handling

stanleyyconsensys:
This PR is adding a error handling on get-starknet

Base on PR #374 , as the Snap is expecting to throw a error that contains the get-starknet error code

therefore on get-starknet side, we should have a method to extract the error and format into a new error with defined message, which will be `createStarkError`

A usage example can be found in RPC `switch-network`


packages/get-starknet/src/utils/object.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Verify if the given property is exist in the object or not.
```
khanti42, line 2:
```suggestion
 * Verify if the given property exists in the object or not.
```

packages/get-starknet/src/utils/object.test.ts
```diff
@@ -0,0 +1,24 @@
...
+    },
+  };
+
+  it.each(['p1.p2.p3'])('returns true if the property "%s" exist in the object', (property: string) => {
+    const result = isset(obj, property);
+
+    expect(result).toBe(true);
+  });
+
+  it('returns false if the property is not exist in the object', () => {
```
khanti42, line 19:
```suggestion
  it('returns false if the property does not exist in the object', () => {
```

packages/get-starknet/src/utils/object.test.ts
```diff
@@ -0,0 +1,24 @@
...
+  const obj = {
+    p1: {
+      p2a: 1,
+      p2: {
+        p3: 4,
+      },
+    },
+  };
+
+  it.each(['p1.p2.p3'])('returns true if the property "%s" exist in the object', (property: string) => {
```
khanti42, line 13:
```suggestion
  it.each(['p1.p2.p3'])('returns true if the property "%s" exists in the object', (property: string) => {
```

packages/get-starknet/src/utils/error.ts
```diff
@@ -0,0 +1,51 @@
...
+  readonly code: number;
+
+  constructor(message: string, errorCode: number) {
+    super(message);
+    this.code = errorCode;
+  }
+}
+
+/**
+ * Create Custom Error base on the given error code to map with the Wallet API error.
```
khanti42, line 35:
```suggestion
 * Create Custom Error based on the given error code to map with the Wallet API error.
```

packages/get-starknet/src/utils/error.test.ts
```diff
@@ -0,0 +1,29 @@
...
+  });
+
+  it('returns default error code and message if the error code is undefined', () => {
+    const error = createStarkError(undefined);
+    expect(error).toBeInstanceOf(CustomError);
+    expect(error.message).toStrictEqual(defaultErrorMessage);
+    expect(error.code).toStrictEqual(defaultErrorCode);
+  });
+
+  it('returns default error code and message if the error code is not exist in the mapping', () => {
```
khanti42, line 23:
```suggestion
  it('returns default error code and message if the error code does not exist in the mapping', () => {
```

khanti42 (changes requested):

khanti42 (changes requested):
As per discussion elsewhere, let's remove object.test.ts and object.ts as we don't use them

packages/get-starknet/src/utils/object.test.ts
stanleyyconsensys: the file removed

packages/get-starknet/src/utils/object.test.ts
stanleyyconsensys: the file removed

packages/get-starknet/src/utils/object.ts
stanleyyconsensys: the file removed

khanti42 (approved):
LGTM

PR #376, opened 2024-10-10
fix: RPC `starkNet_executeTxn` storing in-correct state data if  the params `calls`  is not an array

khanti42:
### PR Description:

This PR addresses the following changes:

1. **Fixture Update:**
   - The `callsExamples.json` fixture has been updated to handle both cases where `calls` can be either an array or a single object. This ensures that the test scenarios are consistent with the real-world behavior where `calls` might be in different formats.
   
   Changes:
   - Previously, `calls` were always treated as an array. The second example in the fixture now provides `calls` as a single object, allowing the tests to cover this scenario.

2. **Test Enhancements:**
   - Refactored the `executeTxn.test.ts` tests to use `it.each` to handle both cases where `calls` could be either an array or a single object.
   - This approach simplifies the test by parameterizing both cases (`Call[]` and `Call`), ensuring that the code is properly tested for different scenarios without duplicating test logic.
   
   Changes:
   - The previous test for `executeTxn` has been replaced with an `it.each` block, covering both the array and object formats of `calls`.
   - The test correctly asserts that the transaction execution behaves as expected, regardless of the format of `calls`.

3. **Code Improvement:**
   - The `executeTxn.ts` logic was updated to handle both array and non-array formats for `calls`. 
   - A new variable `call` was introduced to check if `calls` is an array and, if so, uses the first element; otherwise, it directly uses the object. This avoids potential issues when `calls` is not an array.
   
   Changes:
   - The line `calls[0]` in the `addTransaction` method was replaced with the newly created `call` variable, ensuring correct handling of both formats.
   
### Summary:
This PR ensures that both the code and tests properly handle `calls` when it is an array or a single object, improving robustness and coverage in different scenarios.

### DIFF
--- a/packages/starknet-snap/src/rpcs/executeTxn.test.ts
+++ b/packages/starknet-snap/src/rpcs/executeTxn.test.ts
@@ -74,12 +74,15 @@ const prepareMockExecuteTxn = async (
     address: account.address,
   });
 
+  const createInvokeTxnSpy = jest.spyOn(executeTxn as any, 'createInvokeTxn');
+
   return {
     network: state.networks[0],
     account,
     request,
     confirmDialogSpy,
     createAccountSpy,
+    createInvokeTxnSpy,
     executeTxnRespMock,
     executeTxnUtilSpy,
     getEstimatedFeesSpy,
@@ -88,10 +91,8 @@ const prepareMockExecuteTxn = async (
 };
 
 describe('ExecuteTxn', () => {
-  let callsExample: any;
-
   it('executes transaction correctly if the account is deployed', async () => {
-    callsExample = callsExamples[0];
+    const calls = callsExamples.multipleCalls;
     const {
       account,
       createAccountSpy,
@@ -100,9 +101,9 @@ describe('ExecuteTxn', () => {
       getEstimatedFeesRepsMock,
       request,
     } = await prepareMockExecuteTxn(
-      callsExample.hash,
-      callsExample.calls,
-      callsExample.details,
+      calls.hash,
+      calls.calls,
+      calls.details,
       true,
     );
 
@@ -116,7 +117,7 @@ describe('ExecuteTxn', () => {
       request.calls,
       undefined,
       {
-        ...callsExample.details,
+        ...calls.details,
         maxFee: getEstimatedFeesRepsMock.suggestedMaxFee,
         resourceBounds:
           getEstimatedFeesRepsMock.estimateResults[0].resourceBounds,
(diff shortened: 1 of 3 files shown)


packages/starknet-snap/src/rpcs/executeTxn.ts
```diff
@@ -176,8 +176,11 @@ export class ExecuteTxnRpc extends AccountRpcController<
       throw new Error('Failed to execute transaction');
     }
 
+    // Ensure it's a single call, whether `calls` is an array or not.
```
stanleyyconsensys, line 179:
```suggestion
    // As the input support `calls` in a single `call` object or in a array of `call` object, 
    // and the state's data structure is not supporting multiple `call` object in a single transaction yet, 
    // then we will have to convert the `calls` into a single `call` object as temp fix
```

packages/starknet-snap/src/rpcs/executeTxn.test.ts
```diff
@@ -90,41 +90,53 @@ const prepareMockExecuteTxn = async (
...
-      request,
-    } = await prepareMockExecuteTxn(
-      callsExample.hash,
-      callsExample.calls,
-      callsExample.details,
-      true,
-    );
+  it.each([
+    {
+      calls: callsExamples[0],
```
stanleyyconsensys, line 95: nit:

IMO, the fixture actually do doing in a good format, 
as you have to using index to identify it is array or not array call

i will suggest change the fixture format, by giving some meaningful key on that (please refer to other fixture)

packages/starknet-snap/src/rpcs/executeTxn.test.ts
```diff
@@ -90,41 +90,53 @@ const prepareMockExecuteTxn = async (
...
+      callStr: 'Call[]', // This is 'callStr'
+    },
+    {
+      calls: callsExamples[1],
+      callStr: 'Call', // This is 'callStr'
+    },
+  ])(
+    'executes transaction correctly if the account is deployed and calls is $callStr',
```
stanleyyconsensys, line 103: i think you should create a test case to check the state, as this test is not checking the state

and the `it.each` should apply to that case, but not that one

packages/starknet-snap/src/rpcs/executeTxn.test.ts
```diff
@@ -90,41 +90,53 @@ const prepareMockExecuteTxn = async (
...
+    async ({ calls }: { calls: any }) => {
+      callsExample = calls;
```
stanleyyconsensys, line 105: btw what is the reason to have a var refer to the inputs? 
the test suppose to be stateless

packages/starknet-snap/src/rpcs/executeTxn.test.ts
```diff
@@ -90,41 +90,53 @@ const prepareMockExecuteTxn = async (
...
-    } = await prepareMockExecuteTxn(
-      callsExample.hash,
-      callsExample.calls,
-      callsExample.details,
-      true,
-    );
+  it.each([
+    {
+      calls: callsExamples[0],
+      callStr: 'Call[]', // This is 'callStr'
```
stanleyyconsensys, line 96: what does it mean by `callStr `
if think u are refer to the call object type?

i will prefer

testCaseTitle: 'an array of call object'
testCaseTitle: 'a call object'

stanleyyconsensys (changes requested):

stanleyyconsensys (changes requested):

packages/starknet-snap/src/rpcs/executeTxn.test.ts
```diff
@@ -126,10 +128,61 @@ describe('ExecuteTxn', () => {
     expect(createAccountSpy).not.toHaveBeenCalled();
   });
 
+  it.each([
+    {
+      callsExampleKey: 'multipleCalls',
```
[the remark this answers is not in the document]
stanleyyconsensys: TBH, callsExampleKey is very wield name

and if u have to do
const calls = callsExamples[callsExampleKey];

why not put the calls in here?

stanleyyconsensys (approved):
LGTM, thank you very much for the update

PR #377, opened 2024-10-11
refactor(get-starknet): change `MetaMaskSnapWallet` lock handling

stanleyyconsensys:
This PR is to change the `MetaMaskSnapWallet` lock handling on `init` method

This is because, we no longer need to init with given network,
But we might have a use case that the caller of `init` already running a lock
With the update , we can allow caller to choose init with lock or not

in addtional, this PR also update the test of the MetaMaskSnapWallet to increase the coverage

### DIFF
--- a/packages/get-starknet/src/wallet.ts
+++ b/packages/get-starknet/src/wallet.ts
@@ -11,6 +11,7 @@ import { MetaMaskSigner } from './signer';
 import { MetaMaskSnap } from './snap';
 import type { MetaMaskProvider, Network } from './type';
 import type { IStarknetWalletRpc } from './utils';
+import { WalletRpcError, WalletRpcErrorCode } from './utils/error';
 
 export class MetaMaskSnapWallet implements StarknetWindowObject {
   id: string;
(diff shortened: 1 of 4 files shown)


packages/get-starknet/src/wallet.ts
```diff
@@ -120,37 +128,54 @@ export class MetaMaskSnapWallet implements StarknetWindowObject {
...
-      this.isConnected = true;
-    });
+    this.#network = network;
+    this.#chainId = network.chainId;
+    this.isConnected = true;
   }
 
+  /**
+   * The entry point that trigger by `connect()` in get-starknet.
+   * It initializes the `MetaMaskSnapWallet` object and return the address that collect from Snap.
```
khanti42, line 175:
```suggestion
   * Initializes the `MetaMaskSnapWallet` object and retrieves an array of addresses derived from Snap.
   * Currently, the array contains only one address, but it is returned as an array to 
   * accommodate potential support for multiple addresses in the future.
```

packages/get-starknet/src/wallet.ts
```diff
@@ -120,37 +128,54 @@ export class MetaMaskSnapWallet implements StarknetWindowObject {
     return this.#chainId;
   }
 
-  // This method is to make sure the local state network is sync with the snap,
-  // Hence to set the updated address, account object, provider object and chainId
-  async init(net?: Network) {
+  /**
+   * Initializes the wallet by fetching the network and account information.
+   * and set the network, address, account object and provider object.
```
khanti42, line 133:
```suggestion
   * and sets the network, address, account object and provider object.
```

khanti42 (approved):
nit only

PR #378, opened 2024-10-11
feat(get-starknet): add `wallet_switchStarknetChain` rpc method

stanleyyconsensys:
This PR is adding `wallet_switchStarknetChain` into get-starknet

wallet_switchStarknetChain will override the parent execute , as it has to ensure the it will only execute once at a time from the dapp

### Requirement:

- [ ] #377

### DIFF
--- a/packages/get-starknet/src/rpcs/switch-network.test.ts
+++ b/packages/get-starknet/src/rpcs/switch-network.test.ts
@@ -1,0 +1,62 @@
+import { mockWalletInit, MainnetNetwork, createWallet, SepoliaNetwork } from '../__tests__/helper';
+import { MetaMaskSnap } from '../snap';
+import type { Network } from '../type';
+import { WalletRpcError } from '../utils/error';
+import { WalletSwitchStarknetChain } from './switch-network';
+
+describe('WalletSwitchStarknetChain', () => {
+  const mockSwitchNetwork = (result: boolean) => {
+    const spy = jest.spyOn(MetaMaskSnap.prototype, 'switchNetwork');
+    spy.mockResolvedValue(result);
+    return spy;
+  };
+
+  const prepareSwitchNetwork = (result: boolean, network?: Network) => {
+    const wallet = createWallet();
+    const { initSpy: walletInitSpy } = mockWalletInit({ currentNetwork: network });
+    const switchNetworkSpy = mockSwitchNetwork(result);
+    return {
+      wallet,
+      walletInitSpy,
+      switchNetworkSpy,
+    };
+  };
+
+  it('switchs the network', async () => {
+    const expectedResult = true;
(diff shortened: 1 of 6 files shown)


packages/get-starknet/src/rpcs/switch-network.ts
```diff
@@ -8,21 +8,37 @@ type Params = RpcTypeToMessageMap[WalletSwitchStarknetChainMethod]['params'];
...
 export class WalletSwitchStarknetChain extends StarknetWalletRpc {
-  async handleRequest(param: Params): Promise<Result> {
+  async execute(params: Params): Promise<Result> {
+    // Adding a lock can make sure the switching network process can only process once at a time with get-starknet,
+    // For cross dapp switching network, which already handle by the snap,
+    // Example scenario:
+    // [Rq1] wallet init and send switch network B request to snap at T0
+    // [Rq2] wallet init and send switch network B request to snap at T1 <-- this request will be on hold by the lock
+    // [Rq1] confrim request and network switch to B, assign local chain Id to B at T2
+    // [Rq2] lock release, wallet inited and local chainId is B, which is same as request, so we return true directly at T3
```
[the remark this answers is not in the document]
khanti42: What happens if [Rq2] requires switching to network C instead?

packages/get-starknet/src/rpcs/switch-network.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    const result = await walletSwitchStarknetChain.execute({ chainId: MainnetNetwork.chainId });
+
+    expect(result).toBe(expectedResult);
+    expect(switchNetworkSpy).toHaveBeenCalledWith(MainnetNetwork.chainId);
+    // Init will be called before and after switching the network
+    // because the wallet will be re-initialized after switching the network
+    expect(walletInitSpy).toHaveBeenCalledTimes(2);
+  });
+
+  it('returns true directly if the request network is the same with the current network', async () => {
```
khanti42, line 39:
```suggestion
  it('returns true immediately if the requested network matches the current network', async () => {
```

packages/get-starknet/src/rpcs/switch-network.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    const walletSwitchStarknetChain = new WalletSwitchStarknetChain(wallet);
+    const result = await walletSwitchStarknetChain.execute({ chainId: requestNetwork.chainId });
+
+    expect(switchNetworkSpy).not.toHaveBeenCalled();
+    expect(result).toBe(true);
+    // If the request network is the same with the current network, init will be called once
+    expect(walletInitSpy).toHaveBeenCalledTimes(1);
+  });
+
+  it('throws `WalletRpcError` if switching network failed', async () => {
```
khanti42, line 52: Shall we check the error code, is it Unknown error or Network switch error ?

khanti42 (changes requested):

PR #381, opened 2024-10-17
feat: add RPC `starkNet_getDeploymentData`

stanleyyconsensys:
This PR is to add RPC `starkNet_getDeploymentData` into the snap

as it is designed for get-starknet v4

The purpose to have this API is to
- having the checking to the account, where it should deployed / upgrade, and enforce the user to deployed / upgrade in our dapp
- does not need to depulicate the deployment data in the snap and get-starkent

### DIFF
--- a/packages/starknet-snap/src/rpcs/get-deployment-data.ts
+++ b/packages/starknet-snap/src/rpcs/get-deployment-data.ts
@@ -40,6 +40,7 @@ export type GetDeploymentDataResponse = Infer<
 
 /**
  * The RPC handler to get the deployment data.
+ *
  */
 export class GetDeploymentDataRpc extends AccountRpcController<
   GetDeploymentDataParams,
@@ -67,8 +68,10 @@ export class GetDeploymentDataRpc extends AccountRpcController<
     params: GetDeploymentDataParams,
   ): Promise<GetDeploymentDataResponse> {
     const { address } = params;
+    // Due to AccountRpcController built-in validation,
     // if the account required to force deploy (Cairo 0 with balance), it will alert with a warning dialog.
     // if the account required to force upgrade (Cairo 0 without balance), it will alert with a warning dialog.
+    // hence we can safely assume that the account is Cairo 1 account.
     if (await isAccountDeployed(this.network, address)) {
       throw new AccountAlreadyDeployedError();
     }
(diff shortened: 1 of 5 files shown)


packages/starknet-snap/src/rpcs/get-deployment-data.ts
```diff
@@ -0,0 +1,88 @@
...
+  ): Promise<GetDeploymentDataResponse> {
+    return super.execute(params);
+  }
+
+  protected async handleRequest(
+    params: GetDeploymentDataParams,
+  ): Promise<GetDeploymentDataResponse> {
+    const { address } = params;
+    // if the account required to force deploy (Cairo 0 with balance), it will alert with a warning dialog.
+    // if the account required to force upgrade (Cairo 0 without balance), it will alert with a warning dialog.
```
khanti42, line 71: This is thanks to the AccountRpcController itself that we extend for. But here it looks it refers to the next function. Shall we add this as description of the handleRequest maybe ?

packages/starknet-snap/src/rpcs/get-deployment-data.ts
```diff
@@ -0,0 +1,88 @@
...
+  typeof GetDeploymentDataRequestStruct
+>;
+
+export type GetDeploymentDataResponse = Infer<
+  typeof GetDeploymentDataResponseStruct
+>;
+
+/**
+ * The RPC handler to get the deployment data.
+ */
```
khanti42, line 43:
```suggestion
/**
 * The RPC handler to get the deployment data.
 * Handles force deploy (Cairo 0 with balance) and force upgrade (Cairo 0 without balance),
 * issuing alert in relevant scenarios.
 */
```

khanti42 (changes requested):
nit on comments

packages/starknet-snap/src/rpcs/get-deployment-data.ts
stanleyyconsensys: what does it mean? any example

packages/starknet-snap/src/rpcs/get-deployment-data.ts
stanleyyconsensys: if we add such comment, we have to add it everywhere

khanti42 (approved):

PR #382, opened 2024-10-17
feat(get-starknet): add `wallet_deploymentData` rpc method

stanleyyconsensys:
This PR is to add `wallet_deploymentData` rpc method


packages/get-starknet/src/rpcs/deployment-data.test.ts
```diff
@@ -0,0 +1,25 @@
+import { mockWalletInit, createWallet } from '../__tests__/helper';
+import { MetaMaskSnap } from '../snap';
+import type { DeploymentData } from '../type';
+import { WalletDeploymentData } from './deployment-data';
+
+describe('WalletDeploymentData', () => {
+  it('returns deployment data', async () => {
```
khanti42, line 7: shall we test that the right error is triggered in case account is deployed ?

khanti42 (changes requested):

packages/get-starknet/src/rpcs/deployment-data.test.ts
stanleyyconsensys: it is the responsibility of the snap, we trust the snap response

stanleyyconsensys: and there is a test to test the error from snap convert to correct response, so no need to implement here

khanti42: ok!

khanti42 (approved):

PR #388, opened 2024-10-18
refactor(SNAP): revamp RPC `starkNet_addErc20Token`

stanleyyconsensys:
This PR is to revamp the RPC `starkNet_addErc20Token` to adopt the new RPC architecture

it also decouple some utils in `SnapUtils.ts` into new helper file

- move `isAsciiString` and  `isValidAsciiStrField` to `string.ts`
- move `isPreloadedToken` to `token.ts`


packages/starknet-snap/src/rpcs/watch-asset.test.ts
```diff
@@ -0,0 +1,170 @@
...
+
+    return {
+      getNetworkSpy,
+      confirmDialogSpy,
+      upsertTokenSpy,
+      request,
+    };
+  };
+
+  it('returns true if the token has added', async () => {
```
khanti42, line 92:
```suggestion
  it('returns true if the token is added', async () => {
```

packages/starknet-snap/src/rpcs/watch-asset.test.ts
```diff
@@ -0,0 +1,170 @@
...
+  it('throws `InvalidNetworkError` if the network can not be found', async () => {
+    const { request, getNetworkSpy } = await prepareWatchAssetTest({});
+    getNetworkSpy.mockResolvedValue(null);
+
+    await expect(watchAsset.execute(request)).rejects.toThrow(
+      InvalidNetworkError,
+    );
+  });
+
+  it('throws `TokenIsPreloadedError` if the given token is one of the preloaded token', async () => {
```
khanti42, line 133:
```suggestion
  it('throws `TokenIsPreloadedError` if the given token is one of the preloaded tokens', async () => {
```

packages/starknet-snap/src/rpcs/watch-asset.ts
```diff
@@ -0,0 +1,191 @@
...
+
+  constructor() {
+    super();
+    this.tokenStateMgr = new TokenStateManager();
+    this.networkStateMgr = new NetworkStateManager();
+  }
+
+  /**
+   * Execute the watch asset request handler.
+   * It will promot a dialog to ask user confirmation.
```
khanti42, line 65:
```suggestion
   * It will prompt a dialog to ask user confirmation.
```

packages/starknet-snap/src/rpcs/watch-asset.ts
```diff
@@ -0,0 +1,191 @@
...
+   *
+   * @param params - The parameters of the request.
+   * @param params.tokenAddress - The address of the token to add.
+   * @param params.tokenName - The name of the token to add.
+   * @param params.tokenSymbol - The symbol of the token to add.
+   * @param params.tokenDecimals - The decimals of the token to add.
+   * @param params.chainId - The chain id of the network to switch.
+   * @returns the response of adding a asset in boolean.
```
khanti42, line 73:
```suggestion
   * @returns the response of adding the asset in boolean.
```

packages/starknet-snap/src/utils/string.test.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  it('returns false for a non ASCII string', () => {
+    // non ASCII string
+    expect(isAsciiString('Schönen Tag noch')).toBe(false);
+  });
+});
+
+describe('isValidAsciiStrField', () => {
+  it.each(['hello', 'hello '])(
+    'returns true for an valid ASCII string: %s',
```
khanti42, line 16:
```suggestion
    'returns true for a valid ASCII string: %s',
```

packages/starknet-snap/src/utils/string.ts
```diff
@@ -0,0 +1,20 @@
...
+ * Check if a string is an ASCII string.
+ *
+ * @param str - The string to check.
+ */
+export function isAsciiString(str: string) {
+  return /^[ -~]+$/u.test(str);
+}
+
+/**
+ * Check if a string is a valid ASCII string field.
```
khanti42, line 11:
```suggestion
 * Check if a string is an ASCII string and meets length constraints.
```

packages/starknet-snap/src/utils/string.test.ts
```diff
@@ -0,0 +1,32 @@
...
+  );
+
+  it.each([
+    // invalid length, longer than 10 chars
+    'Have a nice day',
+    // non ASCII string
+    'Schönen',
+    // non ASCII string
+    ' Schönaa ',
+  ])('returns false for a non valid ASCII string: %s', (str: string) => {
```
khanti42, line 29:
```suggestion
  ])('returns false for a string that fails ASCII check or length validation: %s', (str: string) => {
```

packages/starknet-snap/src/utils/superstruct.ts
```diff
@@ -28,7 +28,35 @@ import {
...
+import { isValidAsciiStrField } from './string';
+
+export const TokenNameStruct = refine(
+  string(),
+  'TokenNameStruct',
+  (value: string) => {
+    if (isValidAsciiStrField(value, MAXIMUM_TOKEN_NAME_LENGTH)) {
+      return true;
+    }
+    return `The given token name is invalid, needs to be in ASCII chars, not all spaces, and has length larger than ${MAXIMUM_TOKEN_NAME_LENGTH}`;
```
khanti42, line 46:
```suggestion
    return `The given token name is invalid, needs to be in ASCII chars, not all spaces, and have length smaller than ${MAXIMUM_TOKEN_NAME_LENGTH}`;
```

packages/starknet-snap/src/utils/superstruct.ts
```diff
@@ -28,7 +28,35 @@ import {
...
+);
+
+export const TokenSymbolStruct = refine(
+  string(),
+  'TokenSymbolStruct',
+  (value: string) => {
+    if (isValidAsciiStrField(value, MAXIMUM_TOKEN_SYMBOL_LENGTH)) {
+      return true;
+    }
+    return `The given token symbol is invalid, needs to be in ASCII chars, not all spaces, and has length larger than ${MAXIMUM_TOKEN_SYMBOL_LENGTH}`;
```
khanti42, line 57:
```suggestion
    return `The given token symbol is invalid, it needs to be in ASCII chars, not all spaces, and have length smaller than ${MAXIMUM_TOKEN_SYMBOL_LENGTH}`;
```

packages/starknet-snap/src/utils/token.test.ts
```diff
@@ -0,0 +1,60 @@
...
+    {
+      ...Config.preloadTokens[0],
+      symbol: 'different symbol',
+    },
+    {
+      ...Config.preloadTokens[0],
+      address: '0x12345',
+    },
+  ])(
+    'returns ture if the token is a preloaded token but with different name, symbol, address',
```
khanti42, line 37:
```suggestion
    'returns true if the token is a preloaded token but with different name, symbol, address',
```

packages/starknet-snap/src/utils/token.ts
```diff
@@ -0,0 +1,35 @@
...
+  chainId,
+}: {
+  tokenName: string;
+  tokenSymbol: string;
+  tokenAddress: string;
+  chainId: string;
+}) {
+  const bigIntTokenAddress = BigInt(tokenAddress);
+  return Boolean(
+    Config.preloadTokens.find(
```
khanti42, line 27: Tokens should be uniquely identified by their `address` and `chainId`, not by `name` or `symbol`.
In the ERC20 standard, multiple tokens can share the same `name` or `symbol`, since these are derived from the token contract.
Therefore, when determining token uniqueness, it's essential to rely on `address` and `chainId` alone.

Wdyt ?

khanti42 (changes requested):

packages/starknet-snap/src/utils/token.ts
stanleyyconsensys: i think that make sense, as i copy the previous behaviour, didnt aware that, but that since our UI does not show the token address, so user can add 2 ETH with different address,

may be this is the reason we checking that way

khanti42 (approved):

PR #393, opened 2024-10-19
refactor(get-starknet): remove the snap connection on rpcs that return static information

stanleyyconsensys:
This PR is to remove the snap connection on rpcs that return static information

as they dont necessary to connect to the snap

APIs include:
- wallet_supportedWalletApi
- wallet_supportedSpecs
- wallet_getPermissions


khanti42 (approved):
LGTM!

PR #394, opened 2024-10-19
fix(get starknet): fix incorrect params in RPC `wallet_signTypedData`

stanleyyconsensys:
this PR is to fix the incorrect params in `wallet_signTypedData` when calling the snap API
- it was using chain id as the address
- but it should use the address


khanti42 (approved):
LGTM!

PR #395, opened 2024-10-19
refactor(get-starknet): change get starknet snap params structure

stanleyyconsensys:
This PR is to change the get starknet snap params structure from arguments in array to arguments in object

it is because it will more easy to read and config with default value

And reduce the get current network call from the new get-starknet PRC
- it was called everytime when submit a request to snap
- it will only called if the chain id is undefined 

### Requirements:
Below PRs required to be merged before hand:
- [x] #394

### DIFF
--- a/packages/get-starknet/src/snap.ts
+++ b/packages/get-starknet/src/snap.ts
@@ -35,10 +35,10 @@ export class MetaMaskSnap {
         snapId: this.#snapId,
         request: {
           method: 'starkNet_extractPublicKey',
-          params: {
+          params: await this.#getSnapParams({
             userAddress,
             chainId,
-          },
+          }),
         },
       },
     })) as string;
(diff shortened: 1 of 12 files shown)


packages/get-starknet/src/snap.ts
```diff
@@ -37,137 +37,180 @@ export class MetaMaskSnap {
           method: 'starkNet_extractPublicKey',
           params: {
```
khanti42, line 38: why here we don't call getSnapParams to set chainId if not set ?

khanti42 (changes requested):

packages/get-starknet/src/snap.ts
stanleyyconsensys: good catch

khanti42 (approved):

PR #397, opened 2024-10-21
fix(UI): remove Snap dependency on Wallet UI package

stanleyyconsensys:
This PR remove the Snap dependency on Wallet UI package

BEGIN_COMMIT_OVERRIDE
fix: Remove Snap dependency on Wallet UI package
END_COMMIT_OVERRIDE


khanti42 (approved):
LGTM!

PR #399, opened 2024-10-22
refine(DeclareContract): revamp RPC starkNet_declareContract 

khanti42:
Refines refactor declare contract to allow fee estimation and include deploy. 

Since there are now shared parts between execute-ten and declare-contract a new rpc class has been created. 
However this create circular import issue during testing. 

Hence this PR is in draft for now.

BEGIN_COMMIT_OVERRIDE
chore: Refine RPC starkNet_declareContract to include fee estimation and deploy account if needed.
END_COMMIT_OVERRIDE


khanti42:
Will re-open later.

PR #400, opened 2024-10-22
feat(get-starknet): Integrate `get-starknet` v4

stanleyyconsensys:
This PR is to add support on get-starknet v4

BEGIN_COMMIT_OVERRIDE
feat: Add get-starknet v4 support
END_COMMIT_OVERRIDE


packages/get-starknet/package.json
```diff
@@ -18,7 +18,7 @@
     "prettier": "prettier --write \"src/**/*.ts\"",
     "lint": "eslint 'src/*.{js,ts,tsx}' --max-warnings 0 -f json -o eslint-report.json",
     "lint:fix": "eslint '**/*.{js,ts,tsx}' --fix",
-    "test:unit": ""
+    "test": "jest --passWithNoTests"
```
khanti42, line 21: Maybe we can add a constraint here ?

khanti42 (approved):
LGTM! Many changes :) If nothing breaks we pay each other a beer !

packages/get-starknet/package.json
stanleyyconsensys: remarks: Add coverage when we have full test implemented

PR #402, opened 2024-10-22
chore(get-starknet): remove `Method not supported` exception for "`on` " and "`off`" event handlers

khanti42:
This PR is to fix the on/off event. The signature is updated to match interface. And the throw is removed to allow the use of WalletAccount as per the starknet documentation: http://starknetjs.com/docs/next/guides/walletAccount#select-a-wallet. If we through the WalletAccount object can't be created.

### DIFF
--- a/packages/get-starknet/src/wallet.test.ts
+++ b/packages/get-starknet/src/wallet.test.ts
@@ -140,7 +140,7 @@ describe('MetaMaskSnapWallet', () => {
     it('throws `Method not supported` error', async () => {
       const wallet = createWallet();
 
-      expect(() => wallet.on()).toThrow('Method not supported');
+      expect(() => wallet.on('accountsChanged', jest.fn())).not.toThrow();
     });
   });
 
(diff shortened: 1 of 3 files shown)


packages/get-starknet/src/wallet.test.ts
```diff
@@ -140,15 +140,15 @@ describe('MetaMaskSnapWallet', () => {
     it('throws `Method not supported` error', async () => {
```
stanleyyconsensys, line 140:
```suggestion
    it('does nothing and not throw any error', async () => {
```

packages/get-starknet/src/wallet.test.ts
```diff
@@ -140,15 +140,15 @@ describe('MetaMaskSnapWallet', () => {
...
       const wallet = createWallet();
 
-      expect(() => wallet.on()).toThrow('Method not supported');
+      expect(() => wallet.on('accountsChanged', jest.fn())).not.toThrow();
     });
   });
 
   describe('off', () => {
     it('throws `Method not supported` error', async () => {
```
stanleyyconsensys, line 148:
```suggestion
    it('does nothing and not throw any error', async () => {
```

stanleyyconsensys (approved):
LGTM

PR #404, opened 2024-10-24
refactor(SNAP): add utils to manage snap ui

stanleyyconsensys:
This PR is to adding some utils method to manage the SNAP UI in a centralise file
to be more convenient when we need to replace with JSX component later

it create some reusable method to generate UI

- rowUI : is to create a row component with text field
- addressUI: is to create a row component with address and linked with explorer
- networkUI: is to create a row component with default label `Network`
- headerUI: is to create a heading component
- dividerUI: is to create a divider component
- signerUI: is to create a addressUI with default label `Signer Address`
- jsonDataUI: is to create a row component with Json Serializer

### DIFF
--- a/packages/starknet-snap/src/config.ts
+++ b/packages/starknet-snap/src/config.ts
@@ -1,3 +1,5 @@
+import { constants } from 'starknet';
+
 import type { Erc20Token, Network } from './types/snapState';
 import {
   SnapEnv,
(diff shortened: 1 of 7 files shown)


khanti42 (approved):
LGTM! thank you very much for this change. Will make JSX implementation lot easier.

PR #412, opened 2024-11-05
chore: add transaction request state manager

stanleyyconsensys:
This PR is to add transaction request state manager to prepare the task when using interactive UI for token unit selection when process execute transaction


khanti42 (approved):
LGTM. I will add the accountDeploy boolean field in a later PR.

PR #413, opened 2024-11-05
feat: event handling in get-starknet

khanti42:
### Description

This PR introduces event handling in `get-starknet` by implementing a polling mechanism to detect and handle `accountsChanged` and `networkChanged` events.

#### Changes Introduced:
1. **Event Polling Added**:
   - Implemented a polling mechanism to detect changes in the wallet account's address and network state.
   - The polling ensures the `walletAccount` object remains in sync with the snap state.

2. **Unified Polling Logic**:
   - A single polling loop is reused for both `accountsChanged` and `networkChanged` event handling.
   - Avoids redundant checks and improves efficiency.

3. **Optimized Polling Mechanism**:
   - **Minimum Delay**: Ensures a minimum delay between successive polls to avoid overloading resources.
   - **Maximum Timeout**: Added a timeout for polling operations to prevent hanging requests.
   - **Sequential Execution**: Ensures that a new poll does not begin until the previous one has completed.

4. **Performance Validation**:
   - Tested with 20 browser tabs open concurrently to ensure the mechanism remains reliable and efficient under heavy usage.

#### Purpose:
This ensure that the walletAccount object address and network are in sync with the snap state. As both events are used in the walletAccount object to ensure this: https://github.com/starknet-io/starknet.js/blob/b15fe2d6aa9a5494c510fcac3d6d34fd7e41bf5c/src/wallet/account.ts#L56


packages/get-starknet/src/wallet.ts
```diff
@@ -220,13 +236,108 @@ export class MetaMaskSnapWallet implements StarknetWindowObject {
...
+    const { signal } = this.#pollingController;
+
+    while (!signal.aborted) {
+      const startTime = Date.now(); // Track the start time of the iteration
+      const previousNetwork = this.#network;
+      const previousAddress = this.#latestAddress;
+
+      try {
+        // Perform the polling operation with a timeout
+        await Promise.race([
```
stanleyyconsensys, line 296: combine with what you try to do at the last line of the function (100ms)

i guess what you try do to go to the next scan ASAP, 

i think it is not clear on the function , may be add a bit description or example

packages/get-starknet/src/wallet.ts
```diff
@@ -220,13 +236,108 @@ export class MetaMaskSnapWallet implements StarknetWindowObject {
...
+        if (previousNetwork.chainId !== this.#network.chainId) {
+          this.#networkChangeHandlers.forEach((callback) => callback(this.#network.chainId, [this.#selectedAddress]));
+        }
+
+        // Check for account change
+        if (previousAddress !== this.#selectedAddress) {
+          this.#accountChangeHandlers.forEach((callback) => callback([this.#selectedAddress]));
+        }
+      } catch (error) {
+        if (!signal.aborted) {
```
stanleyyconsensys, line 314: add small comment
```suggestion
        // Not to throw an error, to avoid breaking the event subscription
        if (!signal.aborted) {
```

packages/get-starknet/src/wallet.ts
```diff
@@ -220,13 +236,108 @@ export class MetaMaskSnapWallet implements StarknetWindowObject {
...
+          console.error('Error in polling:', error.message);
+        }
+      }
+
+      // Calculate remaining time for minimum delay
+      const elapsedTime = Date.now() - startTime;
+      const remainingTime = Math.max(0, MetaMaskSnapWallet.minimumPollingDelay - elapsedTime); // Ensure a minimum delay of 100ms
+      if (remainingTime > 0) {
+        await new Promise((resolve) => setTimeout(resolve, remainingTime));
+      }
```
stanleyyconsensys, line 324: always have a small buffer in between each loop, wdyt

```suggestion
     await new Promise((resolve) => setTimeout(resolve, min(MetaMaskSnapWallet.minimumPollingDelay, remainingTime)));

```

packages/get-starknet/src/wallet.ts
```diff
@@ -220,13 +236,108 @@ export class MetaMaskSnapWallet implements StarknetWindowObject {
...
+          new Promise(
+            (_, reject) =>
+              setTimeout(() => reject(new Error('Polling timeout exceeded')), MetaMaskSnapWallet.maximumPollingDelay), // Timeout after 5 seconds
+          ),
+        ]);
+
+        // Check for network change
+        if (previousNetwork.chainId !== this.#network.chainId) {
+          this.#networkChangeHandlers.forEach((callback) => callback(this.#network.chainId, [this.#selectedAddress]));
+        }
```
stanleyyconsensys, line 307: i guess you dont wanna break if any of handler fail
and it also better running in a "async" way ?

assume u have some function like below 
```typescript
const resolver = (func, arg) => {
  return new Promise((resolve) => {  
    resolve(func(arg))
  }); 
}
```
in that case, it can support either async or non-async, wont break and run all function at once
(for better design, please use batch call for max 50 calls in one go, please refer to [here](https://github.com/MetaMask/snap-bitcoin-wallet/blob/main/packages/snap/src/utils/async.ts))



```suggestion
        if (previousNetwork.chainId !== this.#network.chainId) {
          await Promise.allSettled(this.#networkChangeHandlers.map(
               (callback) => resolver(callback, this.#selectedAddress)
          ))
        }
```

packages/get-starknet/src/wallet.ts
```diff
@@ -220,13 +236,108 @@ export class MetaMaskSnapWallet implements StarknetWindowObject {
...
+
+    while (!signal.aborted) {
+      const startTime = Date.now(); // Track the start time of the iteration
+      const previousNetwork = this.#network;
+      const previousAddress = this.#latestAddress;
+
+      try {
+        // Perform the polling operation with a timeout
+        await Promise.race([
+          this.#init(), // Perform the main operation
```
stanleyyconsensys, line 297:
```suggestion
           // Fetch network, assign address and chainId for thread safe.
          this.init(),
```

packages/get-starknet/src/wallet.ts
```diff
@@ -220,13 +236,108 @@ export class MetaMaskSnapWallet implements StarknetWindowObject {
...
+   */
+  #pollingFunction = async (): Promise<void> => {
+    if (!this.#pollingController) {
+      return; // Abort if the polling controller is not initialized
+    }
+    const { signal } = this.#pollingController;
+
+    while (!signal.aborted) {
+      const startTime = Date.now(); // Track the start time of the iteration
+      const previousNetwork = this.#network;
```
stanleyyconsensys, line 291: nit:
we have chain id in the class
```suggestion
      const previousNetwork = this.#chainid;
```

packages/get-starknet/src/wallet.ts
```diff
@@ -220,13 +236,108 @@ export class MetaMaskSnapWallet implements StarknetWindowObject {
...
+      const previousAddress = this.#latestAddress;
```
stanleyyconsensys, line 292: we have selected address, and btw `this.#latestAddress` is never assigned too

```suggestion
      const previousAddress = this.#selectedAddress;
```

stanleyyconsensys (commented):
the overall coding look good,
some advice for clean up the code
and some bug/wrong behaviour discovered:
- any handler error will break other handler

packages/get-starknet/src/wallet.ts
```diff
@@ -26,6 +26,12 @@ import type { MetaMaskProvider, Network } from './type';
 import type { IStarknetWalletRpc } from './utils';
 import { WalletRpcError, WalletRpcErrorCode } from './utils/error';
 
+const resolver = async (func: (...args: any[]) => any, ...args: any[]): Promise<any> => {
```
stanleyyconsensys, line 29: nit:
i guess you know what arg it is , so you dont need any?

packages/get-starknet/src/wallet.ts
```diff
@@ -220,13 +239,107 @@ export class MetaMaskSnapWallet implements StarknetWindowObject {
...
+
+  /**
+   * Un-subscribe the `accountsChanged` or `networkChanged` event for a given handler.
+   *
+   * @param event - The event name ('accountsChanged' or 'networkChanged').
+   * @param handleEvent - The event handler function to un-subscribe.
+   */
+  off<Event extends keyof WalletEventHandlers>(event: Event, handleEvent: WalletEventHandlers[Event]): void {
+    if (event === WalletEvent.AccountsChanged) {
+      this.#accountChangeHandlers = this.#accountChangeHandlers.filter((handler) => handler !== handleEvent);
```
stanleyyconsensys, line 269: did you confirm with get-starknet team on the off handle?

if above is correct
i would suggest using MAP/SET rather than array to store the func

packages/get-starknet/src/wallet.ts
```diff
@@ -220,13 +239,107 @@ export class MetaMaskSnapWallet implements StarknetWindowObject {
...
+
+      await new Promise((resolve) => setTimeout(resolve, MetaMaskSnapWallet.minimumPollingDelay));
+    }
+  };
+
+  #startPolling(): void {
+    this.#pollingController = new AbortController();
+    this.#pollingFunction().catch((_error) => {
+      // Do nothing
+    });
```
stanleyyconsensys, line 336: i think we are safe to let the error not catch
as in the function itself, it has try catch on every scan
but if it has an error outside the while loop, then it is unexpected, we can throw that error
```suggestion
    this.#pollingFunction();
```

packages/get-starknet/src/wallet.ts
```diff
@@ -55,6 +61,16 @@ export class MetaMaskSnapWallet implements StarknetWindowObject {
...
 
+  #pollingController: AbortController | undefined;
+
+  #accountChangeHandlers: AccountChangeEventHandler[] = [];
+
+  #networkChangeHandlers: NetworkChangeEventHandler[] = [];
+
+  static readonly minimumPollingDelay = 100;
+
+  static readonly pollingTimeoutMs = 5000;
```
stanleyyconsensys, line 72: nit:
static var is similar with constants, lets use upper case at the first char (if the lint allow)
```suggestion
  static readonly PollingDelayMs = 100;

  static readonly PollingTimeoutMs = 5000;
```

packages/get-starknet/src/wallet.ts
```diff
@@ -220,13 +239,107 @@ export class MetaMaskSnapWallet implements StarknetWindowObject {
...
+        // If `this.#init()` takes too long, the timeout will reject and continue to the next iteration.
+        await Promise.race([
+          // Fetch network, assign address and chainId for thread safe.
+          this.#init(),
+          new Promise((_, reject) =>
+            // Timeout after `MetaMaskSnapWallet.pollingTimeoutMs`.
+            setTimeout(() => reject(new Error('Polling timeout exceeded')), MetaMaskSnapWallet.pollingTimeoutMs),
+          ),
+        ]);
+
```
stanleyyconsensys, line 308: should we do it in parallel for network change and account change?

so the account change doesnt need to wait all events from network change

packages/get-starknet/src/wallet.ts
khanti42: lint does not allow start with caps but I renamed the constants with your suggestion

stanleyyconsensys (approved):
LGTM, we can have a unit test later on

PR #415, opened 2024-11-06
chore: add jsx support in snap

khanti42:


package.json
```diff
@@ -25,6 +25,7 @@
     "@commitlint/config-conventional": "^17.0.3",
     "@lavamoat/allow-scripts": "^3.0.0",
     "@lavamoat/preinstall-always-fail": "^2.0.0",
+    "@types/react": "18.0.15",
```
stanleyyconsensys, line 28: what is the reason to add this special version? if this is cause by type error, why other snap does not have this issue?

stanleyyconsensys (commented):
overall good, but not sure why types react needed to add

package.json
khanti42: let's remove this.

stanleyyconsensys (approved):
LGTM

PR #416, opened 2024-11-06
feat: add jsx support detection

khanti42:
This PR is to handle MM version that don't support JSX (<v11)

## Requirements 

- [x] : #415


packages/starknet-snap/src/utils/snap.ts
```diff
@@ -93,3 +100,45 @@ export async function setStateData<State>(data: State) {
...
+export const ensureJsxSupport = async (component: Component) => {
+  try {
+    await snap.request({
+      method: 'snap_dialog',
+      params: {
+        type: 'alert',
+        content: component,
+      },
+    });
+    const state = await getStateData<SnapState>();
```
khanti42, line 132: we need to acquire a lock here. Otherwise there could be concurrency issue with the first rpc request from the dApp wouldn't have state.

packages/wallet-ui/src/components/ui/organism/MinVersionModal/MinVersionModal.view.tsx
```diff
@@ -11,21 +11,37 @@ export const MinVersionModalView = () => {
       <StarknetLogo />
       <Title>A new version of the Starknet Snap is available</Title>
```
stanleyyconsensys, line 12: this change should be rollbacked

packages/starknet-snap/src/utils/snap.ts
```diff
@@ -93,3 +98,11 @@ export async function setStateData<State>(data: State) {
     },
   });
 }
+
+export const updateRequiredMetaMaskComponent = () => {
```
stanleyyconsensys, line 102: this should put into the snap-ui.ts

packages/starknet-snap/src/index.tsx
```diff
@@ -302,42 +297,38 @@ export const onRpcRequest: OnRpcRequestHandler = async ({ request }) => {
...
-          <Text>
-            To manage your Starknet account and send and receive funds, visit
-            the <Link href={getDappUrl()}>companion dapp for Starknet</Link>.
-          </Text>
-        </Box>
-      ),
-    },
-  });
+  const initSnapStateManager = new InitSnapStateManager();
+  await initSnapStateManager.ensureJsxSupport(
```
stanleyyconsensys, line 301: IMO, we should use a helper, rather than a state manager, which out of the purpose for managing state

1. Check if JSX supported by the utils
2. if true, then update the state
3. if false, then display a dialog

and i think we dont need to store in state, as it is TEMP (eventually browser should use the latest MM version)

we can just check it in onInstall, onUpdate, onHomepage

for standard request, just let it fail

packages/starknet-snap/src/index.tsx
```diff
@@ -302,42 +297,38 @@ export const onRpcRequest: OnRpcRequestHandler = async ({ request }) => {
...
+        As usual, to manage your Starknet account and send and receive funds,
+        visit the <Link href={getDappUrl()}>companion dapp for Starknet</Link>.
+      </Text>
+    </Box>,
+  );
 };
 
 export const onHomePage: OnHomePageHandler = async () => {
+  const initSnapStateManager = new InitSnapStateManager();
+  const requireMMUpgrade = await initSnapStateManager.requireMetaMaskUpgrade();
```
stanleyyconsensys, line 327: it will cause racing condition, 
homepage and oninstall and onupdate are running at same time

and i think we can do the same with onUpdate and onInstall

stanleyyconsensys (commented):
We dont need a state mgr , as this is TEMP solution, and eventually needed to remove when most of the user has the latest MM version

stanleyyconsensys (changes requested):
the init state mgr still exist
the wallet ui still change when we already split it into other PR

stanleyyconsensys (approved):

PR #418, opened 2024-11-11
feat: add fee token selection interactive UI

khanti42:
## Requirements : 

- [x] : #415 
- [x] : #417

PR #417, opened 2024-11-06
feat: add common jsx component

khanti42:
## Requirements : 

- [x] : #415

PR #419, opened 2024-11-11
feat: adopt fee token selection dialog in RPC `starkNet_executeTxn`

khanti42:
## Requirements : 

- [x] : #418 
- [x] : #417 
- [x] : #415


packages/starknet-snap/src/utils/formatter-utils.ts
```diff
@@ -37,3 +42,67 @@ export const mapDeprecatedParams = <Params>(
...
+      } catch (error) {
+        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+        console.warn(`Error in amount conversion: ${error}`);
+      }
+    }
+
+    dataArray.push(callData);
+  }
+
+  return dataArray;
```
stanleyyconsensys, line 107: change the function name, and remain it not to process multiple, but format a single item

please also update the corresponding comment of the function , include params

```suggestion
export const callToTransactionReqCall = async (
  call: Call,
  chainId: string,
  address: string,
  tokenStateManager: TokenStateManager,
): Promise<FormattedCallData> => {
    const { contractAddress, calldata, entrypoint } = call;

    // Base data object for each call, with transfer fields left as optional
    const formattedCall: FormattedCallData = {
      contractAddress,
      calldata: calldata as string[],
      entrypoint,
      isTransfer: false, // Set default to false
    };

    // Check if the contract is an ERC20 token and entrypoint is 'transfer' to populate transfer fields
    const token = await tokenStateManager.getToken({
      address: contractAddress,
      chainId,
    });

    if (token && entrypoint === 'transfer' && calldata) {
      try {
        const senderAddress = address;
        
        // ensure the data is in correct format, 
        // if an error occur, it will catch and not to format it
        assert(calldata[0], AddressStruct)
        assert(calldata[1], NumberStringStruct)

        const recipientAddress = calldata[0]; // Assuming calldata[0] is the recipient address
        const amount = calldata[1]; // Convert amount using token decimals

        // Populate transfer-specific fields
        formattedCall.isTransfer = true;
        formattedCall.transferSenderAddress = senderAddress;
        formattedCall.transferRecipientAddress = recipientAddress;
        formattedCall.transferAmount = amount;
        formattedCall.transferTokenSymbol = token.symbol;
        formattedCall.transferTokenDecimals = token.decimals;
      } catch (error) {
        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
        logger.warn(`Error in amount conversion: ${error.message}`);
      }
    }
    return formattedCall
```

stanleyyconsensys (commented):
adding some comment on formatCallData

and please also combine some change with the dependency PR

stanleyyconsensys (approved):
LGTM

PR #420, opened 2024-11-12
feat: add event listener for fee token selection

khanti42:
## Description

- A user-input abstract controller class to manage user input in interactive ui 
- An implementation of that class for fee selection 
- Its use in the context of executeTxn rpc call

## Requirements 

- [ ] : #419 
- [ ] : #418 
- [ ] : #417 
- [ ] : #415


packages/starknet-snap/src/index.tsx
```diff
@@ -341,3 +355,112 @@ export const onUpdate: OnUpdateHandler = async () => {
...
+    </Box>,
+  );
+
+  switch (eventKey) {
+    case generateEventKey(
+      UserInputEventType.InputChangeEvent,
+      'feeTokenSelector',
+    ): {
+      const networkStateMgr = new NetworkStateManager();
+      const network = await networkStateMgr.getCurrentNetwork();
```
stanleyyconsensys, line 396: we should use the network from the request, to simplify you can even pass the network object in the request
so you dont need to have chainId & chain name from the transactionRequest

packages/starknet-snap/src/ui/utils.tsx
```diff
@@ -80,3 +92,149 @@ export async function generateExecuteTxnFlow(
...
+  feeToken: FeeToken,
+  suggestedMaxFee: string,
+): Promise<boolean> {
+  const stateManager = new TokenStateManager();
+
+  // Fetch STRK token details and balance
+  const strkToken = await stateManager.getStrkToken({
+    chainId: network.chainId,
+  });
+  const strkBalance = strkToken
```
stanleyyconsensys, line 188: nit:

if we have the token and network information in the context, then we dont need to fetch it here
i remember why you wanna to format the call, because we dont wanna do it over and over

packages/starknet-snap/src/ui/utils.tsx
```diff
@@ -80,3 +92,149 @@ export async function generateExecuteTxnFlow(
...
+ * @param address - The wallet address to check balances for.
+ * @param network - The network object, which includes information such as chainId.
+ * @param calls - An array of contract call objects.
+ * Each object should contain a contract address and an optional amount.
+ * @param feeToken - Specifies the fee token type (either STRK or ETH) to use for the transaction fee.
+ * @param suggestedMaxFee - The maximum transaction fee to consider, as a stringified bigint.
+ * @returns Resolves to `true` if there are sufficient funds for the specified fee token
+ * to cover both the total amounts in `calls` and the `suggestedMaxFee`; otherwise, `false`.
+ */
+export async function hasSufficientFunds(
```
stanleyyconsensys, line 175: look like we can simplify it, instead of doing all if else then for strk/eth

as we only need to consider if the user exceed their balance for 1 fee token
 (if you wanna to check the user has sufficient balance for what he is going to send, is another story, as he can send dai, abc, def)

e.g
```typescript
aysnc function hasSufficientFundsForFee({
  feeToken,
  network,
  address,
  calls,
}: {
  feeToken: Token, 
  network: Network, 
  address:string
  calls: FormattedCallData[]
}) {
    const balanceForFeeToken = await getBalance(
        address,
        feeToken.address,
        network,
        BlockIdentifierEnum.Pending,
      )
    : '0';
    
    // Calculate total STRK or ETH amounts from `calls`
    const totalSpendForFeeToken = calls.reduce((acc, call) => {
      const { tokenTransferData , contractAddress } = call
      if (tokenTransferData && contractAddress === feeToken.address) {
        acc += BigInt(tokenTransferData.amount);
      }
      return acc
    }, BigInt(suggestedMaxFee)) 

    return totalSpendForFeeToken >= balanceForFeeToken;
}
```

packages/starknet-snap/src/rpcs/execute-txn.ts
```diff
@@ -146,8 +146,10 @@ export class ExecuteTxnRpc extends AccountRpcController<
       interfaceId: '',
       type: TransactionType.INVOKE,
       signer: address,
+      addressIndex: this.account.addressIndex,
       maxFee: suggestedMaxFee,
       calls: formattedCalls,
+      resourceBounds: estimateResults.map((result) => result.resourceBounds),
```
stanleyyconsensys, line 152: im not sure if i getting wrong or not, it seem missing to pull back the est result to below for execute txn

stanleyyconsensys (changes requested):
some change request
and looks like missing 2 important features

1. what happen if error, user can still click the confirm
2. what happen after user select a token, it doesnt bring back to execute txn

packages/starknet-snap/src/ui/components/ExecuteTxnUI.tsx
```diff
@@ -89,7 +103,10 @@ export const ExecuteTxnUI: SnapComponent<ExecuteTxnUIProps> = ({
             )}
           </Section>
         ))}
-        <FeeTokenSelector selectedToken={selectedFeeToken as FeeToken} />
+        <FeeTokenSelector
+          selectedToken={selectedFeeToken as FeeToken}
+          error={errors?.fees}
```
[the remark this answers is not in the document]
khanti42: Why not using the default Field error option ? <Field label="Fee Token" error={error}> like we do here ? https://docs.metamask.io/snaps/features/custom-ui/#field

packages/starknet-snap/src/utils/formatter-utils.test.ts
```diff
@@ -65,3 +72,92 @@ describe('mapDeprecatedParams', () => {
     expect(requestParams).toStrictEqual(expected);
   });
 });
+
+describe('callToTransactionReqCall', () => {
```
stanleyyconsensys, line 82: nit:
the test should be more realistic and we can re-use some existing constant

stanleyyconsensys (commented):
Let some comments but i fixed and pushed, if you okay, i will approve after ur review

it looks like we missing a test for the new flow
and missing the test for the controller too

stanleyyconsensys (approved):

PR #422, opened 2024-11-17
feat: replace rpc dialogs with jsx

khanti42:
## Requirements 

- [ ] : #415 
- [ ] : #417


stanleyyconsensys (changes requested):
i remember we discussed we create a component for each UI
and create function for those UI to render
so that the org file doesnt need to change to tsx

stanleyyconsensys (approved):

stanleyyconsensys (approved):
LGTM

stanleyyconsensys (approved):
LGTM

PR #433, opened 2024-11-24
fix: improve Fee Token Selection UX with Disclaimer and Error Messaging

khanti42:
> **This PR addresses a QA request to improve messaging.**

### PR Description:

This PR improves the user experience for fee token selection by adding a default disclaimer and clearer error messaging:

1. Disclaimer: A default message informs users about the fallback mechanism:

`If the chosen token has no funds, the system will automatically switch to a funded token.`

2. Error Messaging: When an unfunded token is selected, the disclaimer is replaced with:

`Not enough STRK to pay for fees. The fee token has been switched back to ETH.`

Why This Change?

- Clarifies fallback behavior to reduce user confusion.
- Enhances feedback during token selection and transaction submission.

BEGIN_COMMIT_OVERRIDE
chore: Improve Fee Token selection UX with better Error Messaging and Disclaimer
END_COMMIT_OVERRIDE


packages/starknet-snap/src/ui/controllers/user-input-event-controller.ts
```diff
@@ -191,8 +191,8 @@ export class UserInputEventController {
     } catch (error) {
       const errorMessage =
         error instanceof InsufficientFundsError
-          ? `Not enough ${feeToken} to pay for fee`
-          : 'Fail to calculate the fees';
+          ? `Not enough ${feeToken} to pay for fee, switching back to ${request.selectedFeeToken}`
```
stanleyyconsensys, line 194: add test

packages/starknet-snap/src/ui/fragments/FeeTokenSelector.tsx
```diff
@@ -47,6 +48,12 @@ export const FeeTokenSelector: SnapComponent<FeeTokenSelectorProps> = ({
           </SelectorOption>
         </Selector>
       </Field>
+      {error ? null : (
```
stanleyyconsensys, line 51: it look weird when the error message is not because of no funds


as i mentioned, it is better to swtich use a separate error component, and display the error msg all in one?
wdyt?

{Boolean(amount.error) && <Text color="error">{amount.error}</Text>}

stanleyyconsensys (commented):
minor comment, mainly missing test

packages/starknet-snap/src/ui/fragments/FeeTokenSelector.tsx
stanleyyconsensys: keep both error message

stanleyyconsensys (approved):

PR #435, opened 2024-11-25
chore: set get-starknet Snap's version through env variable

khanti42:
Version enforcement is now handled globally using `process.env.SNAP_VERSION`. 

Retained the `_snapVersion` parameter in the MetaMaskSnapWallet constructor for backward compatibility, as some existing consumers explicitly pass it.

### DIFF
--- a/packages/get-starknet/src/wallet.ts
+++ b/packages/get-starknet/src/wallet.ts
@@ -64,7 +64,7 @@ export class MetaMaskSnapWallet implements StarknetWindowObject {
   /**
    * Initializes a new instance of the MetaMaskSnapWallet class.
    *
-   * The Snap version is now enforced globally via a static `snapVersion` property, 
+   * The Snap version is now enforced globally via a static `snapVersion` property,
    * this ensures consistent versioning across all instances and removes the need for consumers to specify it.
    *
    * @param metamaskProvider - The MetaMask Wallet Provider.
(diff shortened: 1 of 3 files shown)


packages/get-starknet/src/wallet.ts
```diff
@@ -58,14 +58,27 @@ export class MetaMaskSnapWallet implements StarknetWindowObject {
...
+
+  /**
+   * Initializes a new instance of the MetaMaskSnapWallet class.
+   *
+   * The Snap version is now enforced globally via a static `snapVersion` property:
+   *
+   * This ensures consistent versioning across all instances and removes the need for consumers to specify it.
+   * However, the `_snapVersion` parameter remains to maintain compatibility with existing usage.
+   * @param metamaskProvider
+   * @param _snapVersion
```
stanleyyconsensys, line 72:
```suggestion
   * Initializes a new instance of the MetaMaskSnapWallet class.
   *
   * The Snap version is now enforced globally via a static `snapVersion` property, this ensures consistent versioning across all instances and removes the need for consumers to specify it.
   * 
   * @param metamaskProvider - The MetaMask Wallet Provider.
   * @param _snapVersion - The `_snapVersion` parameter remains to maintain compatibility with existing usage.
```

stanleyyconsensys (commented):
i think you forgot too add it back in webpack

stanleyyconsensys (approved):
LGTM

stanleyyconsensys (approved):

PR #436, opened 2024-11-25
chore: update test for `UserInputEventController`

stanleyyconsensys:
This PR is to update the test in  `UserInputEventController`

and it also fixing  the bug in generate `errorMessage`
- it was using the selectedFeeToken from the updated request, but the request may change, it was passed because the test is not correct
- it using the selectedFeeToken from the original request, and adding a test to make sure both token not same


khanti42 (dismissed):
LGTM!

khanti42 (approved):
LGTM

PR #437, opened 2024-11-25
feat: allow users to transfer money to a .stark name

### LINKED ISSUE #185: feat: allow users to transfer money to a .stark name
This feature will add support to sending money to a .stark domain within the app. Here is what it could look like:

Capture d’écran 2023-11-30 à 15 25 16

**Implementation proposal**

To retrieve the address from the user stark name, we can call the function `domain_to_address` from the starknet ID [naming contract](https://starkscan.co/contract/0x6ac597f8116f886fa1c97a23fa4e08299975ecaf6b598873ca6792b9bbfb678#read-write-contract). This function takes in argument an encoded domain as a felt array and returns the a Starknet address. You can find more information on our encoding and decoding algorithm [here](https://docs.starknet.id/architecture/naming/encoding).

This feature will follow a similar implementation than #184. 

In starknet-snap : 
- Create a new method in [Starknet MetaMask Snap API](https://github.com/Consensys/starknet-snap/blob/main/packages/starknet-snap/openrpc/starknet_snap_api_openrpc.json) : `starkNet_getAddrFromStarkName`

In wallet-ui : 
- add a call to `starkNet_getAddrFromStarkName` and use it [SendModal](https://github.com/Consensys/starknet-snap/blob/main/packages/wallet-ui/src/components/ui/organism/Header/SendModal/SendModal.view.tsx) component
- update the `AddressInput` component and the `handleChange` function to also accept a Stark name 

If a significant number of contributors agree with our approach I'll start working on Pull Request.

Please feel free to answer on this issue and tell me what you think !

irisdv:
This PR adds support to sending money to a .stark domain.

Close #185


stanleyyconsensys (changes requested):
Thank you very much for the work @irisdv 

We did some revamp on the SNAP, and some of the RPC is consist by new structure

for this API, you can refer to the example of ./rpcs/watch-asset.ts

the major changes are :
- we use OOP and inherit from a RpcController
- we use superstruct to validate the input
- we have NetworkStateManager to get network by chain id
- we dont do logging of the response, the base RpcController will do it 
- we use jest, but not mocha

irisdv:
> We did some revamp on the SNAP, and some of the RPC is consist by new structure

Thanks @stanleyyconsensys , I updated my PR

packages/starknet-snap/src/rpcs/get-addr-from-starkname.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  protected async handleRequest(
+    params: GetAddrFromStarkNameParams,
+  ): Promise<GetAddrFromStarkNameResponse> {
+    const { chainId, starkName } = params;
+
+    const network = await this.getNetworkFromChainId(chainId);
+    const getAddressResp = await getAddrFromStarkNameUtil(network, starkName);
+
+    return getAddressResp;
```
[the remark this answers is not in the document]
irisdv: `Account.getAddressFromStarkName()` from starknet.js will throw an error. Should I add a check in here to ensure that the `address` we get from `getAddrFromStarkNameUtil` is an `AddressStruct` ?

packages/wallet-ui/src/components/ui/molecule/AddressInput/AddressInput.view.tsx
```diff
@@ -62,9 +76,27 @@ export const AddressInputView = ({
     if (isValid) {
       setValid(true);
       setError('');
+      onResolvedAddress?.(inputRef.current.value);
+    } else if (isValidStarkName(inputRef.current.value)) {
+      setValid(false);
+      setError('');
+
+      getAddrFromStarkName(inputRef.current.value, chainId).then((address) => {
```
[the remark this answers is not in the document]
irisdv: Indeed, I removed the query

packages/wallet-ui/src/components/ui/organism/Header/SendModal/SendModal.view.tsx
```diff
@@ -64,7 +69,47 @@ export const SendModalView = ({ closeModal }: Props) => {
...
+                if (abortController.signal.aborted) return;
+                if (isValidAddress(address)) {
+                  setResolvedAddress(address);
+                } else {
+                  setResolvedAddress('');
+                  setErrors((prevErrors) => ({
+                    ...prevErrors,
+                    address: '.stark name doesn’t exist',
+                  }));
+                }
```
stanleyyconsensys, line 96: as snap ensured the result, we can make it simple
```suggestion
                setResolvedAddress(address);
```

packages/wallet-ui/src/components/ui/organism/Header/SendModal/SendModal.view.tsx
```diff
@@ -108,8 +180,9 @@ export const SendModalView = ({ closeModal }: Props) => {
             </Network>
             <AddressInput
               label="To"
-              placeholder="Paste recipient address here"
+              placeholder="Paste recipient address or .stark name here"
               onChange={(value) => handleChange('address', value.target.value)}
+              resolvedAddress={resolvedAddress}
```
stanleyyconsensys, line 185: AddressInput is shared with other component, and i saw it has added some logic for checking the starkname internally, so it will mix up the use case

IMO, 
either 
- we build a new AddressInput for that case
or, 
- we rollback and takeout the validate starkname logic from AddressInput, 
- we pass an flag to stop validate anything inside the AddressInput
- we do the validation at parent level, and display the error from parent

packages/wallet-ui/src/components/ui/organism/Header/SendModal/SendModal.view.tsx
```diff
@@ -94,6 +139,33 @@ export const SendModalView = ({ closeModal }: Props) => {
     );
   };
 
+  const getAddrFromStarkNameQuery = async (
+    starkName: string,
+    options?: { signal?: AbortSignal },
+  ): Promise<string> => {
+    const { signal } = options || {};
+
+    return new Promise((resolve, reject) => {
```
stanleyyconsensys, line 148: the abort controller doesnt consolidate the request, but just put them in queue

it means

when i first type
"alex.stark" => call snap once
then i type in 123456789 continually 
"alex12345689.stark" => call 9 time

stanleyyconsensys (commented):
Hi @irisdv , the SNAP side look good to me, thanks for the change

but the UI side may have some issue

packages/wallet-ui/src/components/ui/organism/Header/SendModal/SendModal.view.tsx
```diff
@@ -108,9 +159,15 @@ export const SendModalView = ({ closeModal }: Props) => {
...
+              disableValidate
+              validateError={errors.address}
+              validInput={isValidAddrField()}
             />
+            {isValidStarkName(fields.address) && resolvedAddress && (
+              <InfoText>{shortenAddress(resolvedAddress, 12)}</InfoText>
+            )}
             <SeparatorSmall />
             <MessageAlert
               variant="info"
```
stanleyyconsensys, line 173: please help to update a message below
from 
"Please only enter a valid Starknet address or .stark name. Sending funds to a different network might result in permanent loss."
to 
"Please only enter a valid Starknet address. Sending funds to a different network might result in permanent loss."

packages/wallet-ui/src/components/ui/organism/Header/SendModal/SendModal.view.tsx
```diff
@@ -64,7 +71,41 @@ export const SendModalView = ({ closeModal }: Props) => {
...
+                  }
+                })
+                .catch((error) => {
+                  if (error.name !== 'AbortError') {
+                    setResolvedAddress('');
+                    setErrors((prevErrors) => ({
+                      ...prevErrors,
+                      address: '.stark name doesn’t exist',
+                    }));
+                  }
```
stanleyyconsensys, line 101: this checking can be removed?
```suggestion
     setResolvedAddress('');
       setErrors((prevErrors) => ({
        ...prevErrors,
        address: '.stark name doesn’t exist',
      }));
```

packages/wallet-ui/src/components/ui/molecule/AddressInput/AddressInput.view.tsx
```diff
@@ -22,20 +22,32 @@ import { STARKNET_ADDRESS_LENGTH } from 'utils/constants';
...
 }: Props) => {
   const [focused, setFocused] = useState(false);
   const inputRef = useRef<HTMLInputElement>(null);
   const [error, setError] = useState('');
   const [valid, setValid] = useState(false);
 
+  useEffect(() => {
+    if (!disableValidate || !inputRef.current) return;
+    setValid(validInput ?? false);
+    setError(validateError || '');
```
stanleyyconsensys, line 48:
```suggestion
    setError(validateError ?? '');
```

packages/wallet-ui/src/components/ui/molecule/AddressInput/AddressInput.view.tsx
```diff
@@ -22,20 +22,32 @@ import { STARKNET_ADDRESS_LENGTH } from 'utils/constants';
...
+  validateError,
+  validInput,
   ...otherProps
 }: Props) => {
   const [focused, setFocused] = useState(false);
   const inputRef = useRef<HTMLInputElement>(null);
   const [error, setError] = useState('');
   const [valid, setValid] = useState(false);
 
+  useEffect(() => {
```
stanleyyconsensys, line 45: nit: 
i think we can simplify a bit, so we can get rid of double validation

as valid is true when
- has no error
- value is not empty

the tradeoff is the icon may show "tick" if the fetching is in process, but if the stark name is not correct, it will turn back to "cross"

```typescript
useEffect(() => {
    if (!disableValidate || !inputRef.current) return;
    setValid(inputRef.current.value !== '' && validateError === '');
    setError(validateError ?? '');
  }, [disableValidate, validateError, inputRef]);

```

stanleyyconsensys (commented):
Hi @irisdv thank you very much for your help

there is not bug at all, just some nit comment to simplify the code and enhance the performance

packages/wallet-ui/src/components/ui/organism/Header/SendModal/SendModal.view.tsx
```diff
@@ -64,7 +71,41 @@ export const SendModalView = ({ closeModal }: Props) => {
...
+                .then((address) => {
+                  if (isValidAddress(address)) {
+                    setResolvedAddress(address);
+                  } else {
+                    setResolvedAddress('');
+                    setErrors((prevErrors) => ({
+                      ...prevErrors,
+                      address: '.stark name doesn’t exist',
+                    }));
+                  }
```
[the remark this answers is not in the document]
stanleyyconsensys: no worry, it will be ensure the result is correct by the superstruct by default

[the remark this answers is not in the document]
irisdv: Ah ok, I just updated the code !

stanleyyconsensys (commented):
hi @irisdv , would you mind to do a rebase?

no worry let me do it for you

stanleyyconsensys (approved):

PR #442, opened 2024-11-28
chore: remove legacy API

stanleyyconsensys:
This PR is to deprecate legacy API which does not needed

- starkNet_sendTransaction -> it has replaced by executeTxn
- starkNet_estimateFees -> it has replaced by starkNet_estimateFee which support multiple fee estimate


khanti42 (approved):
LGTM

PR #443, opened 2024-11-28
chore: add base controller `ChainRpcController`

stanleyyconsensys:
This PR is to add `ChainRpcController` for the case that a RPC doesnt need account discover, but just getting the network instance by a chainId 

the `ChainRpcController` will place into './rpcs/abstract'

due to stateManager is included  `ChainRpcController`, it will result a cycle require when some RPC extends the ChainRpcController

and the other abstract RPC controllers should move to the same place


khanti42 (approved):
LGTM

PR #444, opened 2024-11-28
chore: relocate base RPC controllers

stanleyyconsensys:
This PR relocate all base RPC controllers into './rpcs/abstract'

it is becaus:
- align with `ChainRpcController`
- avoid cycle require 

### Requirements
This PR require below PR(s) to be merged:
- [x] #443


packages/starknet-snap/src/rpcs/switch-network.ts
```diff
@@ -3,8 +3,9 @@ import { assign, boolean } from 'superstruct';
 
 import { NetworkStateManager } from '../state/network-state-manager';
 import { renderSwitchNetworkUI } from '../ui/utils';
-import { AuthorizableStruct, BaseRequestStruct, RpcController } from '../utils';
+import { AuthorizableStruct, BaseRequestStruct } from '../utils';
 import { InvalidNetworkError, UserRejectedOpError } from '../utils/exceptions';
+import { RpcController } from './abstract/base-rpc-controller';
```
khanti42, line 8: Why not using chain-rpc-controller here ?

khanti42 (dismissed):
LGTM

khanti42 (approved):

PR #445, opened 2024-11-28
chore: Add Cache-Busting Query Parameters to Webpack Output Files

khanti42:
## PR Description

This PR introduces cache-busting query parameters to Webpack’s filename and chunkFilename configurations. This change ensures that both entry point files and dynamically imported chunks avoid issues caused by DNS or browser caching, particularly in production environments. The addition of query parameters (?v=[fullhash]) helps load the latest versions of assets, even if stale files are cached.


stanleyyconsensys (approved):
LGTM

PR #448, opened 2024-12-03
feat: add permission boundary

stanleyyconsensys:
This PR is to add permission boundary to limit the access of some RPC API by different origins ([reference](https://consensyssoftware.atlassian.net/jira/software/projects/SF/boards/472?selectedIssue=SF-738))

It also include:
- due to no use case exist, we removed the access for RPCs - `starkNet_addNetwork`, `starkNet_estimateAccountDeployFee` and `starkNet_getStoredTransactions`

BEGIN_COMMIT_OVERRIDE
feat: Add permission boundary.  Block RPCs - `starkNet_addNetwork`, `starkNet_estimateAccountDeployFee` and `starkNet_getStoredTransactions` access for any origin (#448)
END_COMMIT_OVERRIDE


khanti42 (approved):
LGTM

PR #450, opened 2024-12-05
chore: refactor `TransactionStateManager` to handle new transaction data structure

stanleyyconsensys:
This PR is to refactor `TransactionStateManager` to handle new transaction data structure.

it include:
- adding dataVersion filter on findTransactions to make sure it only return new version data
- update test helper to only generate transactions that fit to new transaction data structure.

it wont break current snap work flow, as get data from TransactionStateManager is not yet being used in any RPC / function

Requirements:
Below PR(s) has to merged beforehand
- [ ] #341


khanti42 (approved):
LGTM

PR #451, opened 2024-12-05
feat: replace RPC provider from Alchemy to Infura DIN

khanti42:
This PR is to replace the Alchemy RPC provider by Infura DIN one. 

The required changes in the CI/CD workflow will happen elsewhere.

BEGIN_COMMIT_OVERRIDE
feat: Replace RPC provider from Alchemy to Infura DIN (https://github.com/Consensys/starknet-snap/pull/451) 
END_COMMIT_OVERRIDE


packages/starknet-snap/package.json
```diff
@@ -28,8 +28,8 @@
     "serve": "mm-snap serve",
     "start": "mm-snap watch",
     "test": "yarn run test:unit && yarn run cover:report && yarn run jest",
-    "test:unit": "nyc --check-coverage --statements 50 --branches 50 --functions 50 --lines 50 mocha --colors -r ts-node/register --require test/global.ts \"test/**/*.test.ts\"",
-    "test:unit:one": "nyc --check-coverage --statements 50 --branches 50 --functions 50 --lines 50 mocha --colors -r ts-node/register --require test/global.ts"
+    "test:unit": "nyc --check-coverage --statements 40 --branches 40 --functions 40 --lines 40 mocha --colors -r ts-node/register --require test/global.ts \"test/**/*.test.ts\"",
+    "test:unit:one": "nyc --check-coverage --statements 40 --branches 40 --functions 40 --lines 40 mocha --colors -r ts-node/register --require test/global.ts"
```
stanleyyconsensys, line 32: it will fail for 50%?

packages/starknet-snap/src/utils/din.test.ts
```diff
@@ -0,0 +1,23 @@
+import { constants } from 'starknet';
+
+import { getRPCUrl } from './din';
+
+describe('getRPCUrl', () => {
+  it('returns Mainnet RPC URL if chain id is Mainnet', () => {
+    expect(getRPCUrl(constants.StarknetChainId.SN_MAIN)).toBe(
+      'https://starknet-mainnet.infura.io/v3/',
```
stanleyyconsensys, line 8: can we confirm the blastAPI din URL is the same with https://starknet-mainnet.infura.io/v3/

packages/starknet-snap/src/utils/din.ts
```diff
@@ -0,0 +1,23 @@
+import { constants } from 'starknet';
```
stanleyyconsensys, line 1: IMO, we may better re-name the file with rpc-provider rather than din, 
as i think for other network it will be just the same function

and having the secret in config.ts to align others

packages/starknet-snap/src/utils/rpc-provider.test.ts
```diff
@@ -0,0 +1,25 @@
+import { constants } from 'starknet';
+
+import { Config } from '../config';
+import { getRPCUrl } from './rpc-provider';
+
+describe('getRPCUrl', () => {
+  Config.rpcApiKey = 'test';
```
stanleyyconsensys, line 7:
```suggestion
  beforeEach(function () {
    Config.rpcApiKey = 'test';
  });
  afterEach(function () {
    Config.rpcApiKey = undefined
  });
```

reset it, in case you have some unexpected behaviour (the way using before and after is not suggest by MM, but i think it is okay for that case)

stanleyyconsensys (commented):
Overall LGTM, just some nit comment

packages/starknet-snap/src/utils/rpc-provider.test.ts
```diff
@@ -0,0 +1,30 @@
...
+import { getRPCUrl } from './rpc-provider';
+
+describe('getRPCUrl', () => {
+  beforeEach(function () {
+    Config.rpcApiKey = 'API_KEY';
+  });
+  afterEach(function () {
+    Config.rpcApiKey = '';
+  });
+  it('returns Mainnet RPC URL if chain id is Mainnet', () => {
```
stanleyyconsensys, line 13: nit:
```suggestion

  it('returns Mainnet RPC URL if chain id is Mainnet', () => {
```

packages/starknet-snap/src/utils/rpc-provider.ts
```diff
@@ -0,0 +1,18 @@
+import { constants } from 'starknet';
+
+import { Config } from '../config';
+
+/**
+ * Gets the rpc URL for a given Chain ID.
+ *
+ * @param chainId - The Chain ID.
```
stanleyyconsensys, line 8:
```suggestion
 * @param chainId - The Chain ID.
 * @returns The RPC node endpoint of the corresponding chain.
```

stanleyyconsensys (commented):
LGTM, a nit change and a doc update

stanleyyconsensys (approved):

PR #454, opened 2024-12-06
chore: fix permission boundary

stanleyyconsensys:
This PR is to fix permission boundary to include RPC method 'ping'


khanti42 (approved):
LGTM

PR #456, opened 2024-12-09
fix: consolidate `ResourceBounds` sum

stanleyyconsensys:
This PR is to fix `ResourceBounds` that has not sum after estimate the fees

it include:
- add  helper class `ConsolidateFees`  
- sum resourceBounds with `ConsolidateFees`  
- re-work test to simplify mock estimate fees


khanti42 (approved):
LGTM

PR #462, opened 2024-12-12
chore: refine SNAP for StarkScan migration

stanleyyconsensys:
This PR is to refine the SNAP for StarkScan migration

it includes:
- update hardcode value `transfer` and `upgrade` to enum ContractFuncName ,


khanti42 (approved):
LGTM

PR #467, opened 2024-12-17
fix: hex normalization in getTransactions

khanti42:
QA session discovered an issue in the removal of transactions from State due to a misformatting of the transaction hash in State (does not comply with the 66 characters standard of Starknet). This PR solves this.

### DIFF
--- a/packages/starknet-snap/src/chain/transaction-service.ts
+++ b/packages/starknet-snap/src/chain/transaction-service.ts
@@ -3,7 +3,7 @@ import { TransactionFinalityStatus, TransactionType } from 'starknet';
 import { TransactionStateManager } from '../state/transaction-state-manager';
 import type { Network, Transaction, V2Transaction } from '../types/snapState';
 import { TransactionDataVersion } from '../types/snapState';
-import { dayToSec, msToSec } from '../utils';
+import { dayToSec, msToSec, normalizeStarknetHex } from '../utils';
 import type { IDataClient } from './data-client';
 
 export class TransactionService {
@@ -107,7 +107,6 @@ export class TransactionService {
     const transactionsOnState: Transaction[] = [];
     const transactionsToRemove: string[] = [];
     const transactionsOnChainSet = new Set<string>();
-
     for await (const tx of this.getTransactionsOnChain(
       address,
       contractAddress,
@@ -116,13 +115,13 @@ export class TransactionService {
       transactionsOnChain.push(tx);
       transactionsOnChainSet.add(tx.txnHash);
     }
-
     for await (const tx of this.getTransactionsOnState(
       address,
       contractAddress,
     )) {
+      console.log(tx.txnHash);
       // eslint-disable-next-line no-negated-condition
-      if (!transactionsOnChainSet.has(tx.txnHash)) {
+      if (!transactionsOnChainSet.has(normalizeStarknetHex(tx.txnHash))) {
         transactionsOnState.push(tx);
       } else {
         transactionsToRemove.push(tx.txnHash);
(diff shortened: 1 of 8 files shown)


packages/starknet-snap/src/chain/transaction-service.ts
```diff
@@ -116,13 +115,13 @@ export class TransactionService {
       transactionsOnChain.push(tx);
       transactionsOnChainSet.add(tx.txnHash);
     }
-
     for await (const tx of this.getTransactionsOnState(
       address,
       contractAddress,
     )) {
+      console.log(tx.txnHash);
```
stanleyyconsensys, line 122:
```suggestion
```

btw what is the reason to have normalizeStarknetHex? is it the response of the tx hash from provider is less than 66? which is different than the one that we insert into state?

packages/starknet-snap/src/chain/transaction-service.ts
```diff
@@ -107,7 +107,6 @@ export class TransactionService {
     const transactionsOnState: Transaction[] = [];
     const transactionsToRemove: string[] = [];
     const transactionsOnChainSet = new Set<string>();
-
```
stanleyyconsensys, line 110: keep prev space, i add it for easier view

packages/starknet-snap/src/utils/formatter-utils.ts
```diff
@@ -16,6 +16,28 @@ export const hexToString = (hexStr) => {
...
+  }
+
+  const hexWithoutPrefix = hexString.slice(2); // Remove '0x'
+
+  if (hexWithoutPrefix.length > 64) {
+    throw new Error('Invalid input: Hex value is longer than 64 characters');
+  }
+
+  return `0x${hexWithoutPrefix.padStart(64, '0')}`; // Pad to 64 chars + '0x'
+};
```
stanleyyconsensys, line 39: i guess what you try to do is verify if a string is hex string, and its length cant exceed 64

may be just use validateAndParseAddress from starknet.js, it trim, add hex, case in sensitive

```
 * const address = "0x90591d9fa3efc87067d95a643f8455e0b8190eb8cb7bfd39e4fb7571fdf";
 * const result = validateAndParseAddress(address);
 * // result = "0x0000090591d9fa3efc87067d95a643f8455e0b8190eb8cb7bfd39e4fb7571fdf"
 ```

File: packages/starknet-snap/src/chain/transaction-service.ts [hunk 4]
```diff
@@ -116,13 +115,13 @@ export class TransactionService {
       transactionsOnChain.push(tx);
       transactionsOnChainSet.add(tx.txnHash);
```
stanleyyconsensys, line 116: if we do normalizeStarknetHex below, we should also do here to align the standard,

packages/starknet-snap/src/utils/formatter-utils.ts
khanti42: Did not know of validateAndParseAddress will also replace other manual padding we were doing with this

packages/starknet-snap/src/chain/transaction-service.ts
khanti42: I prefer Option 1, as it ensures that the transaction hash format is consistently normalized wherever it’s used. Since we pull data from the provider every time (and do not store it), normalizing at the source does not offer a significant advantage in terms of efficiency or maintainability.

stanleyyconsensys: i think normalizing the source can be having a valid source of truth, as we also want the output to align as well

stanleyyconsensys (dismissed):

stanleyyconsensys (approved):

PR #469, opened 2024-12-19
chore: fix some typos

fudancoder:


stanleyyconsensys:
Thank you for advice, but we have a upcoming release to address those issue

PR #470, opened 2025-01-07
feat: add account service

stanleyyconsensys:
This PR is to add account service skeleton for the revamp purpose of account discovery and support multiple account

The PR will not impact to existing RPC

The changes include:
- add a `AccountService` to manage and derive account.
- add a `AccountContractDiscovery` to discover the account.
- add a `ContractReader` utils to replace callContract method from starknet utils, and consolidate the error into`ContractReadError` and `ContractNotDeployedError`

- add `cairoVersion` to  `AccContract` state
- add new state dataset `removedAccounts` to save the removed account index
- add `upsertAccount`, `getNextIndex` and `removeAccount` method in `AccountStateManager`


packages/starknet-snap/src/wallet/account/contract.ts
```diff
@@ -0,0 +1,189 @@
...
+
+  /**
+   * Determines whether require deploy is needed.
+   * A contract requires a deploy if it is not deployed and has balance.
+   *
+   * @returns A promise that resolves true if the contract requires a deploy, false otherwise.
+   */
+  async isRequireDeploy(): Promise<boolean> {
+    return (
+      !(await this.isDeployed()) && (await this.getEthBalance()) > BigInt(0)
```
khanti42, line 164: Ultimately we should allow deploy with Strk balance as well. Maybe we add this to backlog ?

packages/starknet-snap/src/wallet/account/cairo1.ts
```diff
@@ -0,0 +1,18 @@
+import type { Calldata } from 'starknet';
+import { CallData } from 'starknet';
+
+import { ACCOUNT_CLASS_HASH } from '../../utils/constants';
+import { CairoAccountContract } from './contract';
+
+export class Cairo1Contract extends CairoAccountContract {
```
khanti42, line 7: As we are using ACCOUNT_CLASS_HASH from Argentx maybe we could make this clearer in the name of the class here to prevent confusion, wdyt ?

packages/starknet-snap/src/state/account-state-manager.ts
```diff
@@ -116,6 +119,28 @@ export class AccountStateManager extends StateManager<AccContract> {
...
+            address: data.address,
+            chainId: data.chainId,
+          },
+          state,
+        );
+
+        if (accountInState) {
+          this.updateEntity(accountInState, data);
+        } else {
+          state.accContracts.push(data);
```
khanti42, line 136: shall we use addAccount here ? Otherwise is there still a reason to keep the addAccount method ?

khanti42 (changes requested):

packages/starknet-snap/src/wallet/account/contract.ts
stanleyyconsensys: but if the account is a Cairo 0 account, it can not use STRK, that y we use ETH

packages/starknet-snap/src/wallet/account/cairo1.ts
stanleyyconsensys: any suggestion, not quite sure what does it mean

packages/starknet-snap/src/state/account-state-manager.ts
stanleyyconsensys: i guess we can remove the addAccount? but i will do it from other PR

packages/starknet-snap/src/wallet/account/reader.ts
```diff
@@ -0,0 +1,36 @@
...
+    return hexToString(resp[0]);
+  }
+
+  /**
+   * Get the ETH balance of the account contract.
+   *
+   * @param cairoContract - The `CairoAccountContract` object to get the balance of.
+   * @returns A promise that resolves to the balance of the contract.
+   */
+  async getEthBalance(cairoContract: CairoAccountContract): Promise<bigint> {
```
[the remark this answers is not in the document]
stanleyyconsensys: yes the legacy account can only get ETH

packages/starknet-snap/src/wallet/account/cairo1.ts
khanti42: something like `ACCOUNT_CLASS_HASH_ARGENTX_VX`

khanti42 (approved):

PR #475, opened 2025-01-13
feat: add account UI

stanleyyconsensys:
This PR is to enable multi account in wallet-ui

### Changes:
- it contains a '+' button to add a new account
- when a new account added, it will load the corresponding tokens, balances, and txns
- for details: please see https://consensyssoftware.atlassian.net/browse/SF-764

image

### Requirements:
- [ ] #474

PR #480, opened 2025-01-14
feat: add UI hooks for manage multi accounts

stanleyyconsensys:
This PR is to add UI hooks to enable account management in UI

### Changes:
- refactor hook `useStarknetSnap`:
  - to simplify snap interaction logic with `useSnap` hook
  - remove mute error logic in `readContract` and `getTransactionStatus` as it is not necessary
  - replace `getWalletData` with `initWalletData`
  - add `switchAccount`, `getCurrentAccount`
- fix some bugs in Transactions List, which we should not call the api if the token address is not ready

### requirements:
- [ ] #479
- [ ] #475


packages/wallet-ui/src/services/useStarkNetSnap.ts
```diff
@@ -1060,11 +790,32 @@ export const useStarkNetSnap = () => {
...
+      method: 'starkNet_getCurrentAccount',
+      params: {
+        chainId,
+      },
+    });
+  };
+
+  const switchAccount = async (chainId: string) => {
+    return await invokeSnap<Account>({
+      method: 'starkNet_switchtAccount',
```
khanti42, line 804:
```suggestion
      method: 'starkNet_switchAccount',
```

khanti42: This is solved in https://github.com/Consensys/starknet-snap/pull/482 that adds the UI part for switching account.

khanti42 (approved):

PR #484, opened 2025-01-16
feat: account event change get starknet

khanti42:
https://github.com/user-attachments/assets/1bd1c8c5-9581-41cc-bc20-b44e5b88763b


packages/get-starknet/src/wallet.ts
```diff
@@ -207,9 +207,9 @@ export class MetaMaskSnapWallet implements StarknetWindowObject {
       throw new Error('Unable to find the selected network');
     }
 
+    // address is depends on network, if network changes, address will update
```
stanleyyconsensys, line 210: i think it should change in that way

```typescript
   const address = await this.#getWalletAddress(network.chainId);

   if (!this.#network || network.chainId !== this.#network.chainId) {
      // provider is depends on network.nodeUrl, if network changes, set provider to undefine for reinitialization
      this.#provider = undefined;
      // account is depends on address and provider, if network changes, address will update,
      // hence set account to undefine for reinitialization
      this.#account = undefined;
    }

    if (address !== this.#selectedAddress) {
      // account is depend on address,
      // hence set account to undefine for reinitialization
      this.#account = undefined;
    }

    this.#selectedAddress = address
```

packages/get-starknet/src/wallet.ts
```diff
@@ -137,7 +137,7 @@ export class MetaMaskSnapWallet implements StarknetWindowObject {
   }
 
   async #getWalletAddress(chainId: string) {
-    const accountResponse = await this.snap.recoverDefaultAccount(chainId);
+    const accountResponse = await this.snap.getCurrentAccount({ chainId, fromState: true });
 
     if (!accountResponse?.address) {
       throw new Error('Unable to recover accounts');
```
stanleyyconsensys, line 143:
```suggestion
      throw new Error('Unable to retrieve the wallet account');
```

packages/get-starknet/src/snap.ts
```diff
@@ -239,6 +239,22 @@ export class MetaMaskSnap {
     return result[0];
   }
 
+  async getCurrentAccount({ chainId, fromState }: { chainId?: string; fromState?: boolean }): Promise<AccContract> {
```
stanleyyconsensys, line 242: lets remove the recover account method and the corresponding test

stanleyyconsensys (changes requested):

stanleyyconsensys:
i see that the on / off is never tested in unit test, may be we should add it back, else we can forget forever

packages/starknet-snap/src/utils/permission.ts
```diff
@@ -17,7 +17,7 @@ export enum RpcMethod {
   GetCurrentAccount = 'starkNet_getCurrentAccount',
   ListAccounts = 'starkNet_listAccounts',
 
-  SwitchAccount = 'starkNet_swtichAccount',
+  SwitchAccount = 'starkNet_switchAccount',
```
stanleyyconsensys, line 20: may require a another PR? i remember you did some fix on it, but it seem rollback in the base branch?

stanleyyconsensys (commented):
overall LGTM, left a minor comment

stanleyyconsensys (approved):

PR #490, opened 2025-01-21
chore: add unit test for get-starknet event

khanti42:
This PR is to add the unit test for get-starknet event

### Changes:
- having a `MockMetaMaskSnapWallet` to extends `MetaMaskSnapWallet` and override some properties to allow spy
- adding test `accountChanged / networkChanged` event 
- changing the modifier of properties `pollingTimeoutMs` and `pollingDelayMs` in `MetaMaskSnapWallet` to protected, which to enable `MockMetaMaskSnapWallet` to mock it


packages/get-starknet/src/wallet.ts
```diff
@@ -65,9 +65,11 @@ export class MetaMaskSnapWallet implements StarknetWindowObject {
 
   #pollingController: AbortController | undefined;
 
-  #accountChangeHandlers: Set<AccountChangeEventHandler> = new Set();
+  // eslint-disable-next-line no-restricted-syntax
+  private readonly accountChangeHandlers: Set<AccountChangeEventHandler> = new Set();
```
stanleyyconsensys, line 69: the private is not working anyway, i would suggest change to protected

and since you will need it in test
(MetaMaskSnapWallet as any).xxxxx is not a standard way to do so

usually we extend the class, and override the protected var into public

packages/get-starknet/src/wallet.test.ts
```diff
@@ -128,18 +129,78 @@ describe('MetaMaskSnapWallet', () => {
...
+    });
+  });
+
+  describe('event handling', () => {
+    it('triggers the accountsChanged handler when the event occurs', async () => {
+      const { getCurrentAccountSpy } = mockWalletInit({});
+      getCurrentAccountSpy.mockReset();
+
+      const wallet = createWallet();
+      getCurrentAccountSpy.mockResolvedValueOnce(generateAccount({ address: '0xInitialAddress' }));
```
stanleyyconsensys, line 174: move the result of generateAccount into a var, so we can have clear view, and reuse it
e.g
const initialAddress = generateAccount({ address: '0xInitialAddress' })
const newAddress = generateAccount({ address: '0xNewAddress' })

packages/get-starknet/src/wallet.test.ts
```diff
@@ -128,18 +129,78 @@ describe('MetaMaskSnapWallet', () => {
...
+    it('throws an error for unsupported events', () => {
+      const wallet = createWallet();
+      expect(() => wallet.off('unsupportedEvent' as any, jest.fn())).toThrow('Unsupported event: unsupportedEvent');
+    });
+  });
+
+  describe('event handling', () => {
+    it('triggers the accountsChanged handler when the event occurs', async () => {
+      const { getCurrentAccountSpy } = mockWalletInit({});
+      getCurrentAccountSpy.mockReset();
```
stanleyyconsensys, line 171:
```suggestion
      // Reset mocking of `getCurrentAccount` to allow for different results when triggered multiple times. 
      getCurrentAccountSpy.mockReset();
```

add this comment for better understand

or 

you put the new address as the default mocking address
and you use `mockResolvedValueOnce` to make sure the first call is the  initialAddress

that may be more clear

packages/get-starknet/src/wallet.test.ts
```diff
@@ -128,18 +129,78 @@ describe('MetaMaskSnapWallet', () => {
...
+
+      const wallet = createWallet();
+      getCurrentAccountSpy.mockResolvedValueOnce(generateAccount({ address: '0xInitialAddress' }));
+      (MetaMaskSnapWallet as any).pollingDelayMs = 0;
+      const handler = jest.fn();
+      wallet.on('accountsChanged', handler);
+      getCurrentAccountSpy.mockResolvedValueOnce(generateAccount({ address: '0xNewAddress' }));
```
stanleyyconsensys, line 178: didnt we execute thing after we mock, is that any reason?

packages/get-starknet/src/wallet.test.ts
```diff
@@ -128,18 +129,78 @@ describe('MetaMaskSnapWallet', () => {
...
+
+      await new Promise((resolve) => setTimeout(resolve, 100));
+
+      expect(handler).toHaveBeenCalledWith(['0xNewAddress'], undefined);
```
stanleyyconsensys, line 182: you can determined by the number of time `toHaveBeenNthCalledWith`

stanleyyconsensys (changes requested):

stanleyyconsensys (approved):
LGTM

PR #492, opened 2025-01-22
feat: enable multi discovery strategy

khanti42:
PR Description:
This PR introduces support for multiple account discovery strategies within the same Snap version. Users can specify the strategy via a URL parameter called `accountDiscovery`. The available strategies are:
	•	DEFAULT: The standard discovery behavior.
	•	FORCE_CAIRO_0: Forces discovery using the Cairo 0 strategy.
	•	FORCE_CAIRO_1: Forces discovery using the Cairo 1 strategy.

This enhancement improves flexibility by allowing users to select the desired discovery method in case they got into weird situation with multiple account deployed from the same derivation index but with different Account Contract.

### DIFF
--- a/packages/wallet-ui/src/services/useStarkNetSnap.ts
+++ b/packages/wallet-ui/src/services/useStarkNetSnap.ts
@@ -767,10 +767,12 @@ export const useStarkNetSnap = () => {
   const addNewAccount = async (chainId: string) => {
     dispatch(enableLoadingWithMessage('Adding new account...'));
     try {
+      const urlParams = new URLSearchParams(window.location.search);
       const account = await invokeSnap<Account>({
         method: 'starkNet_addAccount',
         params: {
           chainId,
+          accountDiscoveryType: urlParams.get('accountDiscovery'),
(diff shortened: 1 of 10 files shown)


packages/wallet-ui/src/services/useStarkNetSnap.ts
```diff
@@ -767,10 +767,12 @@ export const useStarkNetSnap = () => {
   const addNewAccount = async (chainId: string) => {
     dispatch(enableLoadingWithMessage('Adding new account...'));
     try {
+      const urlParams = new URLSearchParams(window.location.search);
       const account = await invokeSnap<Account>({
         method: 'starkNet_addAccount',
         params: {
           chainId,
+          accountDiscoveryType: urlParams.get('accountDiscovery'),
```
stanleyyconsensys, line 775: why not do those inject from the invokeSnap function? i guess when force deploy/upgrade, u still need that? or no

stanleyyconsensys (commented):
LGTM

PR #494, opened 2025-01-23
feat: Add RPC `starkNet_toggleAccountVisibility` to toggle the account visibility

stanleyyconsensys:
This PR is add a new RPC `starkNet_toggleAccountVisibility` to toggle the account visibility

### Changes:
- Add RPC `starkNet_toggleAccountVisibility`  and allowlist from the permission (restricted)
- remove `removedAccounts` related code
- when derive a account from `AccountService`, we can pass a json data into the `Account` object, hence it can store it when reflect the correct data when using `serialize`


khanti42 (approved):
LGTM!

PR #495, opened 2025-01-23
feat: wallet-ui translation

khanti42:


packages/wallet-ui/src/components/pages/Home/Home.view.tsx
```diff
@@ -13,22 +14,25 @@ export const HomeView = ({ address }: Props) => {
   );
   const loader = useAppSelector((state) => state.UI.loader);
   const { upgradeModalVisible } = useAppSelector((state) => state.modals);
-
+  const { getTranslator } = useStarkNetSnap();
+  const translate = getTranslator();
```
stanleyyconsensys, line 18: the pattern should be sth like that
```suggestion
  const { translate } = useMultiLanguages();
```

packages/wallet-ui/src/services/useStarkNetSnap.ts
```diff
@@ -107,6 +124,55 @@ export const useStarkNetSnap = () => {
...
+          snapId,
+          request: {
+            method: 'starkNet_getPreferences',
+          },
+        },
+      });
+      setLanguage(userLocale as string);
+      return (await import(`../assets/locales/${userLocale}.json`)).messages;
+    } catch (error) {
+      return (await import(`../assets/locales/en.json`)).messages;
```
stanleyyconsensys, line 154: 1. default lang should be loaded before everything, instead of when error happen,  as it is not snap, we should not need to copy directly from SNAP
e.g you can load it at the top , then when ever you need locate, you try to reference it
```
import defaultLocate from '../assets/locales/en.json';

getLocale = useCallback(() => {
   if (locale) return locale;
   return defaultLocate;
  }, [locale]);

```



3. getUserLocalePreference should just return locate

4. for setLanguage and import locate, you can handle it from loadLocale

packages/wallet-ui/src/services/useStarkNetSnap.ts
```diff
@@ -107,6 +124,55 @@ export const useStarkNetSnap = () => {
...
+  const loadLocale = async () => {
+    try {
+      const userLocale = await getUserLocalePreference();
+      setLocale(userLocale);
+    } catch (error) {
+      console.error('Error loading locale:', error);
+    }
+  };
+
+  const getLanguage = useCallback(() => {
```
stanleyyconsensys, line 136: why it need a function to return a single value

stanleyyconsensys (changes requested):
the solution create some memory leak,

and UI is different with SNAP, we should not use the same way that we did for SNAP

packages/wallet-ui/src/services/useStarkNetSnap.ts
```diff
@@ -107,6 +124,55 @@ export const useStarkNetSnap = () => {
...
+          },
+        },
+      });
+      setLanguage(userLocale as string);
+      return (await import(`../assets/locales/${userLocale}.json`)).messages;
+    } catch (error) {
+      return (await import(`../assets/locales/en.json`)).messages;
+    }
+  };
+  const getTranslator = (): Translator | null => {
```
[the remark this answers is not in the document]
khanti42: we directly return translate now and its moved in a useMultiLanguage separate file.

packages/wallet-ui/src/services/useStarkNetSnap.ts
```diff
@@ -974,8 +1040,16 @@ export const useStarkNetSnap = () => {
     }
   };
 
+  useEffect(() => {
```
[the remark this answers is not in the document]
khanti42: the preferences are loaded only at the init

packages/wallet-ui/src/services/useStarkNetSnap.ts
khanti42: Since not language exist we still need to catch the error. However since English is default loaded we can just catch the error but no need to reimport. Since this method is now only called once I think this works.

packages/wallet-ui/src/services/useStarkNetSnap.ts
khanti42: language and locale have been moved to redux. That way getLanguage does not even need to exist we just read state. And dispatch to state at init both language and locale in case locale are loaded properly.

packages/wallet-ui/src/services/useStarkNetSnap.ts
```diff
@@ -107,6 +124,55 @@ export const useStarkNetSnap = () => {
       });
   };
 
+  const loadLocale = async () => {
```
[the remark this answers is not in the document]
khanti42: there are only two methods left one to load preferences (loadPreferences) located in useStarknetSnap (because it requires the provider) this dispatches to redux state. And one translate method in the useMultiLanguage

packages/wallet-ui/src/assets/locales/en.json
```diff
@@ -0,0 +1,308 @@
...
+      "message": "Account upgraded successfully."
+    },
+    "add": {
+      "message": "ADD"
+    },
+    "addToken": {
+      "message": "Add Token"
+    },
+    "addTokenCapital": {
+      "message": "ADD TOKEN"
```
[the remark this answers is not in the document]
khanti42: we now use .toUpperCase for those cases instead of having Capital

stanleyyconsensys (commented):
left some comments

packages/wallet-ui/src/services/useMultiLanguage.ts
```diff
@@ -0,0 +1,18 @@
+import { useAppSelector } from 'hooks/redux';
+
+export const useMultiLanguage = () => {
+  const { locale } = useAppSelector((state) => state.wallet);
+
```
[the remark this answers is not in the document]
khanti42: This would create circular dependencies as we are now loading it in init Snap which is in useStarkNetSnap. Loading it in App.tsx is tricky because we are not awaiting on snap init and snap needs to be inited for the snap call to get the preference to work. After some tests the easier is still to keep both dispatch of locale and translations in UseStarknetsnap maybe we can revisit this later ? Wdyt ?

packages/wallet-ui/src/services/useStarkNetSnap.ts
```diff
@@ -232,6 +257,7 @@ export const useStarkNetSnap = () => {
       dispatch(setActiveNetwork(idx));
       const chainId = net.chainId;
       await getWalletData(chainId, nets);
+      await loadPreferences();
```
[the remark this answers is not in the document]
khanti42: Tried several other ways to load it the simplest was in init. For now it does not retriever many times. Doing it in App.tsx would involve more work as we are not awaiting the init of the Snap and it can create some non desired recall because of React lifecycle. In init there will be one extra call maybe when adding the account but I think this is acceptable. Wdyt ?

packages/wallet-ui/src/components/ui/organism/DeployModal/DeployModal.view.tsx
```diff
@@ -29,13 +29,16 @@ enum Stage {
...
+  const { translate } = useMultiLanguage();
+
   const [txnHash, setTxnHash] = useState('');
   const [stage, setStage] = useState(Stage.INIT);
   const networks = useAppSelector((state) => state.networks);
   const chainId = networks?.items[networks.activeNetwork]?.chainId;
   const toastr = new Toastr();
 
   const onDeploy = async () => {
+    if (!translate) return;
```
stanleyyconsensys, line 41: no necessary
```suggestion
```

packages/wallet-ui/src/components/ui/organism/Header/SendModal/SendModal.view.tsx
```diff
@@ -49,6 +50,7 @@ export const SendModalView = ({ closeModal }: Props) => {
 
   const handleChange = (fieldName: string, fieldValue: string) => {
     //Check if input amount does not exceed user balance
+    if (!translate) return;
```
stanleyyconsensys, line 53:
```suggestion
```

packages/wallet-ui/src/components/ui/organism/UpgradeModel/UpgradeModel.view.tsx
```diff
@@ -28,13 +28,15 @@ enum Stage {
...
   const { upgradeAccount, waitForAccountUpdate } = useStarkNetSnap();
+  const { translate } = useMultiLanguage();
   const [txnHash, setTxnHash] = useState('');
   const [stage, setStage] = useState(Stage.INIT);
   const networks = useAppSelector((state) => state.networks);
   const chainId = networks?.items[networks.activeNetwork]?.chainId;
   const toastr = new Toastr();
 
   const onUpgrade = async () => {
+    if (!translate) return;
```
stanleyyconsensys, line 39:
```suggestion
```

stanleyyconsensys (approved):
some nit comment, but i give a pass first

khanti42:
solving comments in a fix pr

PR #496, opened 2025-01-24
fix: incorrect switch account RPC endpoint in Wallet-UI

stanleyyconsensys:
This PR is to fix the incorrect switch account RPC endpoint in Wallet-UI

### Change:
- Update switch account RPC endpoint to `starkNet_switchAccount `
- Remove duplicate `initWalletData` when adding a new account
- Disable  `InfoModal` temporarily


khanti42 (approved):
LGTM!

PR #499, opened 2025-01-28
feat: wallet-ui multiple accounts scrollbar

khanti42:
This PR adds the support of infinite account to Wallet-UI 

Changes : 

- The "My Account" is replaced by "Account X" where is X is the account number (accountIndex + 1) 
- The list of accounts now uses the Account Icon and Account name and short address for better UI/UX
- When the account number is higher than 3 a scrollbar appears 

See the video below : 

https://github.com/user-attachments/assets/dfcfe21b-4197-4510-91b3-c21fdcc925d3

### DIFF
--- a/packages/wallet-ui/src/components/ui/organism/AddTokenModal/AddTokenModal.view.tsx
+++ b/packages/wallet-ui/src/components/ui/organism/AddTokenModal/AddTokenModal.view.tsx
@@ -106,7 +106,7 @@ export const AddTokenModalView = ({ closeModal }: Props) => {
                 fields.symbol,
                 parseFloat(fields.decimal),
                 chainId,
-                currentAccount,
+                currentAccount?.address || '0x',
               );
               if (newToken) {
                 setErc20TokenBalance(newToken);
--- a/packages/wallet-ui/src/components/ui/organism/AddTokenModal/AddTokenModal.view.tsx
+++ b/packages/wallet-ui/src/components/ui/organism/AddTokenModal/AddTokenModal.view.tsx
@@ -100,19 +100,21 @@ export const AddTokenModalView = ({ closeModal }: Props) => {
           enabled={enabled}
           onClick={async () => {
             try {
-              const newToken = await addErc20Token(
-                fields.address,
-                fields.name,
-                fields.symbol,
-                parseFloat(fields.decimal),
-                chainId,
-                currentAccount?.address || '0x',
-              );
-              if (newToken) {
-                setErc20TokenBalance(newToken);
-                toastr.success('Token added successfully');
+              if (currentAccount) {
+                const newToken = await addErc20Token(
+                  fields.address,
+                  fields.name,
+                  fields.symbol,
+                  parseFloat(fields.decimal),
+                  chainId,
+                  currentAccount.address,
+                );
+                if (newToken) {
+                  setErc20TokenBalance(newToken);
+                  toastr.success('Token added successfully');
+                }
+                closeModal();
               }
-              closeModal();
             } catch (err) {
               toastr.error('Error while adding token');
             }
--- a/packages/wallet-ui/src/components/ui/organism/SideBar/SideBar.view.tsx
+++ b/packages/wallet-ui/src/components/ui/organism/SideBar/SideBar.view.tsx
@@ -27,14 +27,9 @@ import { useStarkNetSnap } from 'services';
 import { DUMMY_ADDRESS } from 'utils/constants';
 import { PopperTooltip } from 'components/ui/molecule/PopperTooltip';
 
-interface Props {
-  address: string;
-  addressIndex: number;
-}
-
-export const SideBarView = ({ address, addressIndex }: Props) => {
+export const SideBarView = () => {
   const networks = useAppSelector((state) => state.networks);
-  const accounts = useAppSelector((state) => state.wallet.accounts);
+  const currentAccount = useAppSelector((state) => state.wallet.currentAccount);
   const chainId = networks?.items[networks.activeNetwork]?.chainId;
   const [listOverflow, setListOverflow] = useState(false);
   const [infoModalOpen, setInfoModalOpen] = useState(false);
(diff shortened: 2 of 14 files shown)


packages/wallet-ui/src/slices/walletSlice.ts
```diff
@@ -53,15 +55,19 @@ export const walletSlice = createSlice({
     ) => {
       const accountsToInsert = Array.isArray(payload) ? payload : [payload];
 
-      const accountSet = new Set(state.accounts);
+      const accountSet = new Set(
+        state.accounts.map((account) => account.address),
+      ); // Create a set of addresses
```
stanleyyconsensys, line 60:
```suggestion
      const accountSet = new Set(
        state.accounts.map((account) => account.address),
      );
```
this is super obvious it is creating a set of addresses

packages/wallet-ui/src/components/ui/organism/SideBar/SideBar.view.tsx
```diff
@@ -113,13 +114,14 @@ export const SideBarView = ({ address }: Props) => {
...
-      <AccountLabel>My account</AccountLabel>
+      <AccountLabel>Account {addressIndex + 1} </AccountLabel>
       <RowDiv>
         <InfoIcon onClick={() => setInfoModalOpen(true)}>i</InfoIcon>
         <AccountSwitchModal
           currentAddress={address}
           starkName={starkName}
-          accounts={accounts}
+          accounts={accounts.map((account) => account.address)}
+          accountsIndex={accounts.map((account) => account.addressIndex)}
```
stanleyyconsensys, line 124: if you already update the slice from storing the account address to the whole account object
then why you need to do 2 pass to loop the accounts to get the address and index, seem not make sense

```suggestion
          accounts={accounts)}
```

packages/wallet-ui/src/slices/walletSlice.ts
```diff
@@ -53,15 +55,19 @@ export const walletSlice = createSlice({
...
       for (const account of accountsToInsert) {
         if (!accountSet.has(account.address)) {
-          state.accounts.push(account.address);
+          state.accounts.push(account);
         }
       }
     },
     setCurrentAccount: (state, { payload }: { payload: Account }) => {
       state.currentAccount = payload.address;
+      state.currentAccountIndex = payload.addressIndex;
```
stanleyyconsensys, line 70:
```suggestion
      state.currentAccount = payload
```
nit:
i would suggest align with setAccounts

but there will be some more files (5 files) to change

but for long term this is good

as you may need more account information later (e.g account name..etc), if you keep adding new props to the component, seem is not for a good plan

stanleyyconsensys (changes requested):
Left some comments, and there are 2 important comments/questions here 
1. Didnt we agree to use modal , rather than pull out to support scroll bar? as. modal can be more responsive, but im okay with that too (Network poll up is different, it only have 2, not dynamic)

2. Address index is a requirement for next ticket (https://consensyssoftware.atlassian.net/browse/SF-785), what is the reason to have this in this ticket? as the user story has not mention this (https://consensyssoftware.atlassian.net/jira/software/projects/SF/boards/472?selectedIssue=SF-784)

khanti42:
The reason to use index here already, is to show Account Number in the list. We could have a separate index but as we already have it from backend I thought it would make sense to add it here already. 

Regarding the modal. I felt that the UX/UI was good. The switch appears closer to the button so it's less extra movement for the mouse and this feels more like an integrated approach. Except if there is a strong no go on that I would suggest to keep it this way. Maybe we can revisit this when integrating design change to include dApps listing..

stanleyyconsensys (changes requested):
Left 2 comments, 

And there is a recommend comment
i can see some component retrieve state from redux, but at the meanwhile it retrieve the redux state from parent

if you dont mind, i will suggest, to align to use redux 
the redux purpose is to avoid pass state from each level

as i already seen passing state from
home -> sidebar -> account detail / account switch

in general a component only stateless if it is a re-useable component , such as AddressInput, AmountInput

i know it may have a lot of change if we do that

I would suggest lets move that slowly, at least
- AccountSwitchModalView -> keep full , and starkName
- SideBarView - no props
- AccountDetailsModal - no props

those can be stateful without the need passing address / addressIndex

in additional, i think you can also make the currentAccount from redux state to have a dummy object to include the dummy address, so when changing SideBarView, you dont need to much effort

WDYT

packages/wallet-ui/src/components/ui/organism/AddTokenModal/AddTokenModal.view.tsx
```diff
@@ -106,7 +106,7 @@ export const AddTokenModalView = ({ closeModal }: Props) => {
                 fields.symbol,
                 parseFloat(fields.decimal),
                 chainId,
-                currentAccount,
+                currentAccount?.address || '0x',
```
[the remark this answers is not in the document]
khanti42: currentAccount can be undefined. 

For example in App.tsx we do `const address = currentAccount?.address ?? DUMMY_ADDRESS;` so two possibilities
- check currentAccount in onClick
- set a DUMMY_ADDRESS like in App.tsx 

I am going for the first one like we do in useEffect. Hope this works. Wdyt ?

khanti42:
Regarding this comment : 

in additional, i think you can also make the currentAccount from redux state to have a dummy object to include the dummy address, so when changing SideBarView, you dont need to much effort

Its solved in an other PR: https://github.com/Consensys/starknet-snap/pull/502

packages/wallet-ui/src/components/ui/organism/AddTokenModal/AddTokenModal.view.tsx
```diff
@@ -100,19 +100,21 @@ export const AddTokenModalView = ({ closeModal }: Props) => {
...
-                fields.name,
-                fields.symbol,
-                parseFloat(fields.decimal),
-                chainId,
-                currentAccount,
-              );
-              if (newToken) {
-                setErc20TokenBalance(newToken);
-                toastr.success('Token added successfully');
+              if (currentAccount) {
```
stanleyyconsensys, line 103: nit: 
not sure if we need to check `if (currentAccount)` , as you will have dummy current account later
and `closeModal();` should execute even if currentAccount not exist?

stanleyyconsensys (approved):
1 comments, but i give a pass first

packages/wallet-ui/src/components/ui/organism/AddTokenModal/AddTokenModal.view.tsx
stanleyyconsensys: i saw that you have already resolve on later branch, which should be fine

PR #500, opened 2025-01-28
chore: wallet UI remove redundant condition check

khanti42:
This PR is to remove a redundant condition check with `translate` on wallet UI


stanleyyconsensys (changes requested):
it looks like the PR is not ready

stanleyyconsensys (approved):
LGTM

PR #501, opened 2025-01-29
feat: wallet-ui multiple accounts hide/unhide feature

khanti42:
This PR adds a hide/unhide feature. 

1. When an active account is hidden it switches to the next not hidden account if available otherwise switch to the first unhidden account 
2. When an non active account is hidden there is no switching of account 
3. When a hidden account is unhide it switches automatically to that account 
4. The hidden accounts can be shown by clicking on the hidden account button in the scrollbar 

See video below for details:

https://github.com/user-attachments/assets/0ebff7b8-c4e7-4a44-8311-818f68a06069


packages/wallet-ui/src/services/useStarkNetSnap.ts
```diff
@@ -191,6 +192,52 @@ export const useStarkNetSnap = () => {
...
+            `Showing account ${shortenAddress(address)}`,
+          ),
+        );
+      }
+      await toggleAccountVisibility(chainId, address, true);
+      dispatch(updateAccount({ address, updates: { visibility: true } }));
+      await switchAccount(chainId, address);
+    } catch (error) {
+      //eslint-disable-next-line no-console
+      console.log(`error while processing unHideAccount: ${error}`);
```
stanleyyconsensys, line 235: you dont need that, as invokeSnap log the error already
```suggestion
```

packages/wallet-ui/src/services/useStarkNetSnap.ts
```diff
@@ -191,6 +192,52 @@ export const useStarkNetSnap = () => {
...
+      const account = await toggleAccountVisibility(chainId, address, false);
+      dispatch(updateAccount({ address, updates: { visibility: false } }));
+      if (account.address !== currentAddress) {
+        await switchAccount(chainId, account.address);
+      }
+    } catch (error) {
+      const toastr = new Toastr();
+      //eslint-disable-next-line no-console
+      console.log(`error while processing hideAccount: ${error}`);
+      toastr.error('You cannot hide the last remaining account.');
```
stanleyyconsensys, line 215: the error is not always means `cannot hide the last remaining account`

packages/wallet-ui/src/services/useStarkNetSnap.ts
```diff
@@ -191,6 +192,52 @@ export const useStarkNetSnap = () => {
...
+      dispatch(disableLoading());
+    }
+  };
+
+  const unHideAccount = async (chainId: string, address: string) => {
+    try {
+      if (!loader.isLoading) {
+        dispatch(
+          enableLoadingWithMessage(
+            `Showing account ${shortenAddress(address)}`,
```
stanleyyconsensys, line 226:
```suggestion
            `Loading...`,
```
nit:
may be just loading is good enough

packages/wallet-ui/src/components/ui/molecule/AccountSwitchModal/AccountSwitchModal.view.tsx
```diff
@@ -17,79 +19,199 @@ import { theme } from 'theme/default';
...
+  useEffect(() => {
+    console.log('Updated accounts:', accounts);
+  }, [accounts]);
+
+  const visibleAccounts = accounts.filter(
+    (account) => account.visibility !== false,
+  );
+  const hiddenAccounts = accounts.filter(
+    (account) => account.visibility === false,
+  );
```
stanleyyconsensys, line 55:
```suggestion
   const visibleAccounts:Account[] = []
   const hiddenAccounts:Account[] = []
   for (const account in accounts) {
       // account.visibility = `undefined` refer to the case when previous account state doesnt include this field
       // hence we consider it is `visible`
       if (account.visibility === undefined || account.visibility === true) {
          visibleAccounts.append(account)
       } else {
          hiddenAccounts.append(account)
       }
   }
```
1 pass and state more clear on undefined case

packages/wallet-ui/src/components/ui/molecule/AccountSwitchModal/AccountSwitchModal.view.tsx
```diff
@@ -17,79 +19,199 @@ import { theme } from 'theme/default';
...
-            const isSelected = account === currentAddress; // Check if the account is selected
-            return (
-              <Menu.Item key={account}>
+
+        {/* Toggle Between Visible and Hidden Accounts */}
+        {showHiddenAccounts ? (
+          /* Hidden Accounts */
+          <MenuSection style={{ height: 201, overflowY: 'auto' }}>
+            {hiddenAccounts.map((account) => (
+              <Menu.Item key={account.address}>
```
stanleyyconsensys, line 112: decouple the visible account and hidden account into a component

packages/wallet-ui/src/components/ui/molecule/AccountSwitchModal/AccountSwitchModal.view.tsx
```diff
@@ -17,79 +19,199 @@ import { theme } from 'theme/default';
...
-                      : 'transparent', // Change background color if selected
-                    borderLeft: isSelected
-                      ? `4px solid ${theme.palette.secondary.main}`
-                      : 'none', // Add left border if selected
-                    paddingLeft: isSelected ? 15 : 20, // Add some padding if selected to make space for the border
+                    paddingLeft: 22,
+                    display: 'flex',
+                    justifyContent: 'space-between',
+                    alignItems: 'center',
+                    opacity: 0.5, // Make it semi-transparent
```
stanleyyconsensys, line 119: nit:
i would suggest move them to the style file

packages/wallet-ui/src/components/ui/molecule/AccountSwitchModal/AccountSwitchModal.view.tsx
```diff
@@ -17,79 +19,199 @@ import { theme } from 'theme/default';
...
+                  }}
+                  onClick={() => {
+                    setShowHiddenAccounts(false);
```
stanleyyconsensys, line 122: unHideAccount will renew the state of accounts, which cause it re-render 
seem setShowHiddenAccounts is useless

packages/wallet-ui/src/components/ui/molecule/AccountSwitchModal/AccountSwitchModal.view.tsx
```diff
@@ -17,79 +19,199 @@ import { theme } from 'theme/default';
...
+                    <IconButton
+                      onClick={(e) => {
+                        if (
+                          account.address !== currentAddress &&
+                          visibleAccounts.length > 2
+                        ) {
+                          e.preventDefault();
+                        }
+                        hideAccount(chainId, account.address, currentAddress);
+                      }}
```
stanleyyconsensys, line 203: e.preventDefault(); should always present, isnt it?

and try move all logics into a function `onHideAccountClick` , apply to all place in this file

stanleyyconsensys (changes requested):
the AccountSwitchModal seem contains many render and logic, i would suggest move them into a small sub component for easier manage

packages/wallet-ui/src/components/ui/molecule/AccountSwitchModal/AccountSwitchModal.view.tsx
khanti42: this is required the re-render of accounts does not reset the showHiddenAccounts state.

packages/wallet-ui/src/components/ui/molecule/AccountSwitchModal/AccountSwitchModal.view.tsx
khanti42: this allows the following UX : 

- Hiding accounts does not close the switch dropdown if there are still hidden accounts left. This allows to hide several accounts more effectively. This is the purpose of e.preventDefault (it prevents the dropdown from closing). See video above @43 seconds 
- If hiding the last possible account to hide (e.g. after hiding only 1 account remains) then the dropdown is closed.

The more general question is when do we want to close or not close the modal. 

Currently 

-> More than two accounts visible, hiding non active accounts does not close the dropdown 
-> All other actions close the dropdown. 

Wdyt ?

packages/wallet-ui/src/components/ui/organism/AccountSwitchModal/HiddenAccountsList/HiddenAccountsList.view.tsx
```diff
@@ -0,0 +1,51 @@
...
+}: Props) => (
+  <MenuSection style={{ height: 201, overflowY: 'auto' }}>
+    {accounts.map((account) => (
+      <Menu.Item key={account.address}>
+        <AccountSwitchMenuItem
+          style={{ paddingLeft: 22, opacity: 0.5 }}
+          onClick={(e) => {
+            setShowHiddenAccounts(false);
+            unHideAccount({ chainId, address: account.address });
+            e.preventDefault();
```
stanleyyconsensys, line 34: you can lets this view focus on render, for event click, you can have a function come from parent
and e.preventDefault(); go first usually
wdyt?
```suggestion
            e.preventDefault();
            onAccountVisibleClick(account);
```

packages/wallet-ui/src/components/ui/organism/AccountSwitchModal/VisibleAccountsList/VisibleAccountsList.view.tsx
```diff
@@ -0,0 +1,89 @@
...
+              paddingLeft: isSelected ? 15 : 22,
+            }}
+          >
+            <AccountSwitchMenuItem
+              style={{ padding: '0px' }}
+              onClick={
+                !isSelected
+                  ? () => switchAccount(chainId, account.address)
+                  : undefined
+              }
```
stanleyyconsensys, line 55: same comment a HiddenAccountList

you can pass in the function

```suggestion
              onClick={onAccountSwitchClick(account)}
```

packages/wallet-ui/src/components/ui/organism/AccountSwitchModal/VisibleAccountsList/VisibleAccountsList.view.tsx
```diff
@@ -0,0 +1,89 @@
...
+                  e.preventDefault();
+                }
+                hideAccount({
+                  chainId,
+                  address: account.address,
+                  currentAddress,
+                });
+              }}
+              size="small"
+            >
```
stanleyyconsensys, line 81: same as above

```suggestion
            <IconButton
              onClick={(e) => {
                // Hiding accounts does not close the switch dropdown if there are still hidden accounts left. 
                // This allows to hide several accounts more effectively. 
                if (accounts.length > 2) {
                  e.preventDefault();
                }
                onAccountHiddenClick(account)
              }}
              size="small"
            >
```

packages/wallet-ui/src/services/useStarkNetSnap.ts
```diff
@@ -202,10 +253,6 @@ export const useStarkNetSnap = () => {
       dispatch(setAccounts(accounts));
     }
 
-    // FIXME: hardcode to set the info modal visible,
-    // but it should only visible when the account is not deployed
-    // dispatch(setInfoModalVisible(true));
```
stanleyyconsensys, line 207: i think better to keep those comment for later we can have the deploy status?

packages/wallet-ui/src/components/ui/organism/AccountSwitchModal/AccountSwitchModal.view.tsx
```diff
@@ -0,0 +1,105 @@
...
+}: Props) => {
+  const networks = useAppSelector((state) => state.networks);
+  const accounts = useAppSelector((state) => state.wallet.accounts);
+  const { switchAccount, addNewAccount, hideAccount, unHideAccount } =
+    useStarkNetSnap();
+  const chainId = networks?.items[networks.activeNetwork]?.chainId;
+
+  const [showHiddenAccounts, setShowHiddenAccounts] = useState(false);
+
+  const visibleAccounts = accounts.filter(
```
stanleyyconsensys, line 37: i think you miss to change some previous comment, please re-visit the hidden comments (unresolved/ hidden)

stanleyyconsensys (changes requested):
thanks for the change, few more comments left

packages/wallet-ui/src/services/useStarkNetSnap.ts
khanti42: sonar was complaining that it does not allow FIXME mentions

packages/wallet-ui/src/services/useStarkNetSnap.ts
khanti42: I moved the toastr in the switchModal and checks if the number of visible Account is < 2 then we show the toast and don't even send the request to the Snap, wdyt ?

packages/wallet-ui/src/components/ui/organism/AccountSwitchModal/AccountSwitchModal.view.tsx
khanti42: should be good now, thanks for the thorough feedback!

packages/wallet-ui/src/services/useStarkNetSnap.ts
```diff
@@ -191,6 +192,52 @@ export const useStarkNetSnap = () => {
...
+      dispatch(updateAccount({ address, updates: { visibility: false } }));
+      if (account.address !== currentAddress) {
+        await initWalletData({
+          account,
+          chainId,
+        });
+      }
+    } catch (error) {
+      //eslint-disable-next-line no-console
+      console.log(`error while processing hideAccount: ${error}`);
```
stanleyyconsensys, line 220: he invokeSnap will log for you, instead may be throw an error and let the consumer to have a toast message to tell the user hiding is fail? or you can directly toast from here

```suggestion
      console.log(`error while processing hideAccount: ${error}`);
```

packages/wallet-ui/src/services/useStarkNetSnap.ts
```diff
@@ -191,6 +192,52 @@ export const useStarkNetSnap = () => {
...
+  }: {
+    chainId: string;
+    address: string;
+  }) => {
+    if (!loader.isLoading) {
+      dispatch(enableLoadingWithMessage(`Loading...`));
+    }
+    await toggleAccountVisibility(chainId, address, true);
+    dispatch(updateAccount({ address, updates: { visibility: true } }));
+    dispatch(disableLoading());
```
stanleyyconsensys, line 238:
```suggestion
    if (!loader.isLoading) {
      dispatch(enableLoadingWithMessage(`Loading...`));
    }
    try {
       await toggleAccountVisibility(chainId, address, true);
       dispatch(updateAccount({ address, updates: { visibility: true } }));
    } catch(err) {
       // or toast the error
       throw err
    } finally {
       dispatch(disableLoading());
    }
```

stanleyyconsensys (changes requested):
thanks for the work, it looks great, 2 minor comment left on the error handle

packages/wallet-ui/src/services/useStarkNetSnap.ts
khanti42: let's directly toast from here!

stanleyyconsensys (approved):
LGTM, thanks for the work

PR #502, opened 2025-02-04
chore: currentAccount is always defined in walletSlice redux state

khanti42:
This PR is to remove the many use of DUMMY_ADDRESS throughout the code base by using a dummy account at the redux state level.

### DIFF
--- a/packages/wallet-ui/src/slices/walletSlice.ts
+++ b/packages/wallet-ui/src/slices/walletSlice.ts
@@ -3,13 +3,14 @@ import { Account } from 'types';
 import { Erc20TokenBalance } from 'types';
 import { Transaction } from 'types';
 import { ethers } from 'ethers';
+import { DUMMY_ADDRESS } from 'utils/constants';
 
 export interface WalletState {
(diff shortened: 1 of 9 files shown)


packages/wallet-ui/src/slices/walletSlice.ts
```diff
@@ -22,7 +23,10 @@ const initialState: WalletState = {
   isLoading: false,
   forceReconnect: false,
   accounts: [],
-  currentAccount: undefined,
+  currentAccount: {
+    address: DUMMY_ADDRESS,
+    addressIndex: 0,
+  } as Account,
```
stanleyyconsensys, line 29: you can have a default account, so that you can reuse, same comment as the line on clearAccounts
```
const defaultAccount = {
    address: DUMMY_ADDRESS,
    addressIndex: 0,
} as Account
```

```suggestion
  currentAccount: defaultAccount,
```

stanleyyconsensys (approved):
LGTM, just 1 nit comment, but i give a pass first

PR #511, opened 2025-02-13
chore: refine account service

stanleyyconsensys:
This PR is to refine the account service and fix some edge case bug

### Changes:
- It will be not longer auto update the account state from `deriveAccountByIndex`, instead the update process split into the workflow that using this method:
  - Update the account state after derive account by address to make sure it sync the data
  - Add the account to state when adding a new account
  - Add the account to state when get the current account if the current account not yet set
- Add validation to make sure the state can only store 1 address per account index 
- remove `upsertAccount` from AccountStateManager to replace by more specific purpose method -`updateAccountByAddress`,`addAccount`

### DIFF
--- a/packages/starknet-snap/src/config.ts
+++ b/packages/starknet-snap/src/config.ts
@@ -38,7 +38,7 @@ export type SnapConfig = {
   };
   account: {
     defaultAccountIndex: number;
-  }
+  };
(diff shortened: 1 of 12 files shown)


khanti42 (approved):
LGTM

PR #512, opened 2025-02-13
chore: add missing translation

khanti42:
This PR is to fix the missing translation from UI that cause by dynamic transaction name from `getTxnName`

### DIFF
--- a/packages/wallet-ui/src/assets/locales/fr.json
+++ b/packages/wallet-ui/src/assets/locales/fr.json
@@ -74,10 +74,10 @@
       "message": "Supprimer la version actuelle dans MetaMask en allant dans Paramètres > Snaps > @consensys/starknet-snap > Voir les détails > Supprimer le Snap"
     },
     "deploy": {
-      "message": "Déployer"
+      "message": "Déploiement"
     },
     "deployAccount": {
-      "message": "Déployer le compte"
+      "message": "Déploiement du compte"
     },
     "deployAccountFailed": {
       "message": "Échec du déploiement du compte"
@@ -230,7 +230,7 @@
       "message": "Sélectionner un jeton pour les frais de transaction"
     },
     "send": {
-      "message": "Envoyer"
+      "message": "Envoie"
     },
     "snapsExtendMetaMask": {
       "message": "Les Snaps étendent les capacités de MetaMask en ajoutant de nouvelles fonctionnalités. Ce Snap permet à MetaMask de être compatible avec Starknet et de gérer vos clés."
--- a/packages/wallet-ui/src/assets/locales/fr.json
+++ b/packages/wallet-ui/src/assets/locales/fr.json
@@ -58,6 +58,9 @@
     "contractAddress": {
       "message": "Adresse du contrat"
     },
+    "contractInteraction": {
+      "message": "Intéraction avec un contrat"
+    },
     "copied": {
       "message": "Copié !"
     },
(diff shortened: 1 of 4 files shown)


packages/wallet-ui/src/assets/locales/fr.json
```diff
@@ -58,6 +58,9 @@
     "contractAddress": {
       "message": "Adresse du contrat"
     },
+    "contractInteraction": {
```
stanleyyconsensys, line 61: where is this key, i dont see we are using that?

stanleyyconsensys (changes requested):

packages/wallet-ui/src/assets/locales/fr.json
khanti42: indeed it was not clear. I simplified the usage to make this crystal clear.

packages/wallet-ui/src/components/ui/molecule/TransactionsList/TransactionListItem/types.ts
```diff
@@ -20,7 +20,8 @@ export const getIcon = (transactionName: string): IconProp => {
 export const getTxnName = (
```
stanleyyconsensys, line 20: i would suggest return the transaction key for this function, instead of doing the transaction internally, which seem a bit messy

you can rename it to getTxnNameKey?

and you should also modify those using the  method that using the name to key

i know we may try to not change too much, but if we keep doing that, some day we may fall into a situration that everything getting more messy, wdyt?

stanleyyconsensys (changes requested):

packages/wallet-ui/src/components/ui/molecule/TransactionsList/TransactionListItem/types.ts
khanti42: my point was that its very hard to track translate(aVariable) then you don't know if you miss a translation. And it makes it very easy to forget some (like I did) this is why I propose the strategy to always call translate('string') so that we know exactly what needs to be translated and can check than all translation exists. Wdyt ?

packages/wallet-ui/src/components/ui/molecule/TransactionsList/TransactionListItem/types.ts
```diff
@@ -3,12 +3,19 @@ import { IconProp } from '@fortawesome/fontawesome-svg-core';
...
+  UpgradeAccount = 'upgradeAccount',
+  ContractInteraction = 'contractInteraction',
+  Deploy = 'deploy',
+  DeployAccount = 'deployAccount',
+  Unknown = 'unknown',
+}
+
 export const getIcon = (transactionName: string): IconProp => {
   switch (transactionName) {
     case 'Send':
```
stanleyyconsensys, line 17: you can change to enum

packages/wallet-ui/src/components/ui/molecule/TransactionsList/TransactionListItem/types.ts
```diff
@@ -17,10 +24,30 @@ export const getIcon = (transactionName: string): IconProp => {
...
+    case TxnType.Deploy:
+      return translate('deploy');
+    case TxnType.DeployAccount:
+      return translate('deployAccount');
+    default:
+      return translate('unknown');
+  }
+};
+
 export const getTxnName = (
```
stanleyyconsensys, line 47: we are return the type enum
```suggestion
export const getTxnType = (
```

stanleyyconsensys (commented):
2 comments

stanleyyconsensys (approved):
LGTM

PR #514, opened 2025-02-14
fix: enhance performance by muting unnecessary state subscription in UI

stanleyyconsensys:
This PR is to fix the redux state on UI components to only subscribe the state that it needed, 

to avoid updating a state and cause all component re-render

e.g 
- at T1 Component A subscribes the wallet state, however it only use 1 of the props - 'xx' from the wallet state;
```
const wallet = useAppSelector((state) => state.wallet)
const varToUse = wallet.xx
```
- at T2, we update a props - 'yy' on wallet
- at T3, Component A will be re-render, which not expected

### Change:
- change `useAppSelector((state) => state.wallet)` to `useAppSelector((state) => state.wallet.{xxxxx}` to only subscribe the state that it needed

### DIFF
--- a/packages/wallet-ui/src/App.tsx
+++ b/packages/wallet-ui/src/App.tsx
@@ -37,7 +37,7 @@ function App() {
   } = useAppSelector((state) => state.modals);
   const { loader } = useAppSelector((state) => state.UI);
   const networks = useAppSelector((state) => state.networks);
-  const { currentAccount } = useAppSelector((state) => state.wallet);
+  const currentAccount = useAppSelector((state) => state.wallet.currentAccount);
   const { hasMetamask } = useHasMetamask();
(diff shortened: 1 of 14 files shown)


khanti42:
thanks for the work. Seems some update are missing in App.tsx, Home.view.tsx, and useStarknetSnap.ts, see here:  https://github.com/Consensys/starknet-snap/compare/chore/fix-ui-redux-state-over-listen...chore/refine-fix-ui-redux-state-over-listen

khanti42 (changes requested):
thanks for the work. Seems some update are missing in App.tsx, Home.view.tsx, and useStarknetSnap.ts, see here: https://github.com/Consensys/starknet-snap/compare/chore/fix-ui-redux-state-over-listen...chore/refine-fix-ui-redux-state-over-listen

khanti42 (approved):

PR #516, opened 2025-02-18
feat: add account drawer UI for account switch

stanleyyconsensys:
This PR is to add Account Drawer UI to replace AccountSwitch pull menu, which enable the end user switch an account from a larger interface

### Change
- Add `AccountDrawer` modal components to show the visible and hidden list of accounts
- Enable end user to switch, hide, show  accounts on the `AccountDrawer` modal components 
- Add reusable component `Scrollable`  for scroll bar view
- Remove SwitchAccount components

### Requirement
- [x] #514


khanti42 (changes requested):
Thanks for the work—this looks great! Just one minor comment: when the last hidden account is `unhidden` the interface should automatically switch to display the `visible` accounts. WDYT?

stanleyyconsensys:
> unhidden

The change is easy, but i found that, i may getting a bit lost after all hidden account visible, because it suddenly appear some accounts after i remove the last hidden account

so i think for now , we keep it like that, and may require a designer to refine that UX

khanti42:
> > unhidden
> 
> The change is easy, but i found that, i may getting a bit lost after all hidden account visible, because it suddenly appear some accounts after i remove the last hidden account
> 
> so i think for now , we keep it like that, and may require a designer to refine that UX , wdyt?

I see could we then maybe have a text just saying no hidden account instead oƒ empty screen

stanleyyconsensys:
> saying

i think that is better

khanti42 (approved):
LGTM!

PR #517, opened 2025-02-21
chore: split MM min version modal

stanleyyconsensys:
This PR is to split the the Modal for `Min Metamask Version` from `MinVersionModal`


khanti42 (approved):
LGTM

PR #518, opened 2025-02-21
fix: the SNAP does not auto update if the version is outdate in wallet UI

stanleyyconsensys:
This PR is to add a fix for a case that the SNAP does not auto update if the version is outdate in wallet UI 

if the SNAP version is outdate , the wallet UI will trigger a reconnect workflow to update the SNAP

### Changes:
- Update the logic to detect the SNAP version is outdate or not
- Does not show an re-install modal when snap is outdate, instead add reconnect workflow 

### Requirements:
 - [x] #517

### DIFF
--- a/packages/wallet-ui/src/App.tsx
+++ b/packages/wallet-ui/src/App.tsx
@@ -82,11 +82,11 @@ function App() {
         <PopIn isOpen={!connected && !loading} showClose={false}>
           <NoMetamaskModal />
         </PopIn>
-        {/* This Modal will be showed when the SNAP must re-installed due to breaking change from Metamask  */}
+        {/* This Modal will be shown when the SNAP must re-installed due to breaking change from Metamask  */}
         <PopIn isOpen={minVersionModalVisible} showClose={false}>
           <MinVersionModal />
         </PopIn>
-        {/* This Modal will be showed when the Metamask version is outdate to support the SNAP */}
+        {/* This Modal will be shown when the Metamask version is outdate to support the SNAP */}
         <PopIn isOpen={minMMVersionModalVisible} showClose={false}>
           <MinMetamaskVersionModal />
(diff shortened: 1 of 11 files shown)


packages/wallet-ui/src/App.tsx
```diff
@@ -81,9 +82,11 @@ function App() {
         <PopIn isOpen={!connected && !loading} showClose={false}>
           <NoMetamaskModal />
         </PopIn>
+        {/* This Modal will be showed when the SNAP must re-installed due to breaking change from Metamask  */}
```
khanti42, line 85:
```suggestion
        {/* This Modal will be shown when the SNAP must re-installed due to breaking change from Metamask  */}
```

packages/wallet-ui/src/App.tsx
```diff
@@ -81,9 +82,11 @@ function App() {
...
         <PopIn isOpen={minVersionModalVisible} showClose={false}>
           <MinVersionModal />
         </PopIn>
+        {/* This Modal will be showed when the Metamask version is outdate to support the SNAP */}
```
khanti42, line 89:
```suggestion
        {/* This Modal will be shown when the Metamask version is outdate to support the SNAP */}
```

khanti42 (changes requested):
nit comments. 
Also we should not forget to create the REACT_APP_MIN_SNAP_VERSION env in the GitHub vars

stanleyyconsensys:
> nit comments. Also we should not forget to create the REACT_APP_MIN_SNAP_VERSION env in the GitHub vars

For now, i guess we dont need to create one, as we always align wanna user to update, means the end user need to use the updated version

but we can create one in other PR with the change on the CICD

khanti42 (approved):

PR #520, opened 2025-02-21
fix: the SNAP does not auto update if the version is outdate in get-starknet

stanleyyconsensys:
This PR is to add a fix for a case that the SNAP does not auto update if the version is outdate in get-starknet

if the SNAP version is outdate , the  get-starknet request workflow will trigger a reconnect workflow to update the SNAP

Changes:
- Add the logic to detect the SNAP version is outdate or not, and trigger re-install

### DIFF
--- a/packages/get-starknet/src/snap.ts
+++ b/packages/get-starknet/src/snap.ts
@@ -24,12 +24,25 @@ import type {
 } from './type';
 
 export class MetaMaskSnap {
+  /**
+   * The wallet RPC provider.
+   */
   #provider: MetaMaskProvider;
 
+  /**
+   * The ID of the SNAP.
+   */
   #snapId: string;
 
+  /**
+   * The version of the SNAP to install.
+   */
   #version: string;
 
+  /**
+   * The minimum version of the SNAP.
+   * If the installed version is lower than this version, the SNAP will be updated.
+   */
   #minSnapVersion: string;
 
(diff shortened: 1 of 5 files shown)


packages/get-starknet/src/snap.ts
```diff
@@ -22,10 +30,14 @@ export class MetaMaskSnap {
...
   #version: string;
 
-  constructor(snapId: string, version: string, provider: MetaMaskProvider) {
+  #minSnapVersion: string;
+
+  constructor(snapId: string, version: string, provider: MetaMaskProvider, minSnapVersion?: string) {
     this.#provider = provider;
     this.#snapId = snapId;
     this.#version = version;
+    // unless specified, the minSnapVersion is the same as the snapVersion
```
khanti42, line 39: nit: Maybe we should be clearer on when minSnapVersion should be used ? E.g. that by default only the latest allowlisted version will be allowed (e.g. all version will have to upgrade) if one wants to have 2 versions working concurrently, e.g. have version that don't force upgrade, then we should use minSnapVersion.

khanti42 (changes requested):
only one nit comment

packages/get-starknet/src/snap.ts
stanleyyconsensys: added a comment into the variable itself

khanti42 (approved):

PR #522, opened 2025-02-26
chore(wallet-ui): fix edge case when account name is undefined in `AccountDetailsModal`

stanleyyconsensys:
This PR is to fix the edge case on wallet UI, 

While edit the name from account detail, but the name is undefined due to previous snap was installed

there will be a JS error caused by accountName.trim()


khanti42 (approved):
LGTM!

PR #525, opened 2025-02-27
feat: add isDeployed field in account state snap

khanti42:
This PR is to add isDeployed field in account state Snap.


packages/starknet-snap/src/wallet/account/account.ts
```diff
@@ -97,6 +97,7 @@ export class Account {
       addressIndex: this.hdIndex,
       chainId: this.chainId,
       cairoVersion: this.cairoVersion,
+      isDeployed: await this.accountContract.isDeployed(),
```
stanleyyconsensys, line 100: nit: 

we can put it all together on await Promise.all()

and do we wanna to store it in the state? as i dont see it has added

And in that case, when you list accounts, you wont have those data, as there is no default data added too

stanleyyconsensys (changes requested):
Overall LGTM, but it seem missing from state, we may need to consider add it back

packages/starknet-snap/src/wallet/account/account.ts
khanti42: The isDeployed is in the `AccContract` type and in the superstruct validation. Is this enough as the getAccount returns the account casting AccContract from the serialise method that does return the isDeployed state.

khanti42: is this enough or do we need to do something else ? Wdyt @stanleyyconsensys

stanleyyconsensys: i think it is ok to not put that, and we can always add it back it we need, resolve the conflict then i will approve

stanleyyconsensys (approved):
LGTM

PR #527, opened 2025-03-03
chore(wallet-ui): fix snap upgrade workflow

stanleyyconsensys:
This PR is to fix the SNAP upgrade workflow

### Changes:
- Remove min snap version from .env.production, as it will impact the deployment
- Adding force upgrade modal for UI, in case the SNAP cant be upgraded for some reason
- Adding upgrade workflow on min snap version modal


packages/wallet-ui/src/components/ui/organism/MinVersionModal/MinVersionModal.view.tsx
```diff
@@ -1,28 +1,28 @@
...
-        <Description>
-          {translate('installLatestVersion')}
-          <ul>
-            <li>{translate('deleteCurrentVersionMetaMask')}</li>
-            <li>{translate('refreshPage')}</li>
-            <li>{translate('connectToMetaMask')}</li>
-          </ul>
-          {translate('accountRecoveryInfo')}
-        </Description>
+
```
khanti42, line 19: We should maybe keep a small description 

```suggestion
<Description>
          {translate('upgradeLatestVersion')}
</Description>
```

And then adding 

en.json
```json
    "upgradeLatestVersion": {
      "message": "To use this dapp, please upgrade to the latest version by clicking the link below:"
    },
```

fr.json
```json
      "upgradeLatestVersion": {
        "message": "Pour utiliser cette dApp, veuillez mettre à jour vers la dernière version en cliquant sur le lien ci-dessous :"
      }
```

wdyt ?

khanti42 (changes requested):
one minor comment

packages/wallet-ui/src/components/ui/organism/MinVersionModal/MinVersionModal.view.tsx
stanleyyconsensys: good idea, just change a bit , instead of link, but button

khanti42 (approved):
LGTM!

PR #528, opened 2025-03-04
chore(wallet-ui): add shared modal style

stanleyyconsensys:
This PR is to add shared modal style for all modals (except Send Transaction & MinVersion, as they are still in other PR)

### Changes
- Add `Modal` component with `Modal.Title`, `Modal.Body`, `Modal.Buttons`, `Modal.Button`, `Modal.Logo` to share in different use case, example:
```
<Modal>
  <Modal.Logo>Title</Modal.Logo>
  <Modal.Title>Title</Modal.Title>
  <Modal.Body>Body</Modal.Body>
  <Modal.Buttons>
       <Modal.Button onClick={} />
       <Modal.Button onClick={} />
   </Modal.Buttons>
</Modal>
```

- Add `Logo` component with Variant for different logo rendering, example:
```
<Logo variant={LogoVariant.Starknet} mr="small" ml="small" mb="small" mt="small" />
```

- Apply `Modal` to all modals (except Send Transaction & MinVersion)

PR #529, opened 2025-03-06
feat(wallet-ui): Enhance account visibility management operation by persisting the account visibility on local

stanleyyconsensys:
This PR is to enhance the UI performance for account visibility management operation

With current implementation, it require an loading screen for account visibility management, which create a non-smooth UX

as the "Visibility" status is not begin use in other DAPP, and likely not in future

Therefore, it is better to stored in UI with the native UI persistent storage (window local storage)

### Changes:
- Add persistent state in UI to store account visibility
- Update `useAccountVisibility` hook to manage the account visibility operation
- Add `useAccountList` hook to get the account list for current chain

### Requirements:
- [ ] #528


khanti42 (approved):
LGTM

PR #532, opened 2025-03-12
feat: prevent account with same name in Snap

khanti42:
This PR is to ensure accountName uniqueness in Snap.
Uniqueness in wallet-ui is here: #535


packages/starknet-snap/src/state/account-state-manager.ts
```diff
@@ -172,6 +184,16 @@ export class AccountStateManager extends StateManager<AccContract> {
...
+        if (
+          account.accountName &&
+          (await this.isAccountNameExist(
+            { accountName: account.accountName, chainId: account.chainId },
+            state,
+          ))
+        ) {
+          throw new Error(`Account name already exists`);
+        }
+
```
stanleyyconsensys, line 196: not sure if we trim b4 or not, may be better to add a safe guard?

```suggestion
        const trimedAccountName = accountName?.trim();
        if (
          trimedAccountName &&
          (await this.isAccountNameExist(
            { accountName: trimedAccountName, chainId },
            state,
          ))
        ) {
          throw new Error(`Account name already exists`);
        }

```

packages/starknet-snap/src/state/account-state-manager.ts
```diff
@@ -435,6 +457,18 @@ export class AccountStateManager extends StateManager<AccContract> {
...
 
+  async isAccountNameExist(
+    { accountName, chainId }: { accountName: string; chainId: string },
+    state,
+  ): Promise<boolean> {
+    const snapState = state ?? (await this.get());
+    return (
+      this.getCollection(snapState).find(
+        (data) => data.accountName === accountName && data.chainId === chainId,
+      ) !== undefined
```
stanleyyconsensys, line 468: same comment as above

```suggestion
      this.getCollection(snapState).find(
        (data) => data.accountName?.trim() === accountName && data.chainId === chainId,
      ) !== undefined
```

stanleyyconsensys: btw does it has a handing on UI?

stanleyyconsensys (commented):
i saw that the trim is not applied, may i ask why?

stanleyyconsensys (approved):
LGTM

PR #533, opened 2025-03-12
fix: invalid inputs on send transaction cause Error

khanti42:
This PR enhances the input validation logic in the SendModal component for the “amount” field:
- Improved validation logic:
- Prevents non-numeric or malformed inputs from triggering errors later in execution.
- Catches parsing errors via ethers.utils.parseUnits and shows a user-friendly "Invalid amount" error message.

New i18n key added:
- invalidAmount message was added to both en.json ("Invalid amount") and fr.json ("Montant invalide").


packages/wallet-ui/src/components/ui/organism/Header/SendInputModal/SendInputModal.view.tsx
```diff
@@ -94,25 +94,36 @@ export const SendInputModalView = ({
...
       case 'amount':
-        if (fieldValue !== '' && fieldValue !== '.') {
-          try {
-            const inputAmount = ethers.utils.parseUnits(
-              fieldValue,
-              erc20TokenBalanceSelected.decimals,
-            );
-            const userBalance = erc20TokenBalanceSelected.amount;
-            if (inputAmount.gt(userBalance)) {
+        if (fieldValue === '' || !isNaN(Number(fieldValue))) {
```
stanleyyconsensys, line 97: im not sure if it is expected

you will set error when

- fieldValue is empty or 
- when fieldValue is not NAN 
  - !isNaN(Number(fieldValue)) => "1.2" => true
  - !isNaN(Number(fieldValue)) => ".2" => true
  - !isNaN(Number(fieldValue)) => "2" => true

stanleyyconsensys (changes requested):

packages/wallet-ui/src/components/ui/organism/Header/SendInputModal/SendInputModal.view.tsx
khanti42: Thanks for the feedback! The behavior is actually intended.
- We don’t show an error when the field is empty — not because it’s valid, but because it’s not a formatting error. The field is marked as required elsewhere in the form logic (so the user can’t proceed with it empty anyway).
- This condition helps us clear any previous error when the user deletes their input — for example, if they typed an invalid value and then removed it, we don’t want to keep showing a red error under the field.
- We enter this logic block when the field is either empty or contains a number. If the field is not empty and not a number, that’s when we show the invalidAmount error.

stanleyyconsensys: i was saying set error 
but i get it, below blocks means set no error for amount field, right
```
setErrors((prevErrors) => ({
            ...prevErrors,
            amount: '',
          }));
```

perhaps it seem better to add description on ur code base, it is not easy to understand what is going to do

i guess u are try to

reset the error if the case is
- empty
- valid input -> !isNaN(Number(fieldValue)) refer to valid input or?

packages/wallet-ui/src/components/ui/organism/Header/SendInputModal/SendInputModal.view.tsx
```diff
@@ -108,6 +116,7 @@ export const SendInputModalView = ({
               }));
             }
           } catch (error) {
+            console.log(error);
```
stanleyyconsensys, line 119: remove console.log

packages/wallet-ui/src/components/ui/organism/Header/SendInputModal/SendInputModal.view.tsx
```diff
@@ -94,7 +94,15 @@ export const SendInputModalView = ({
...
+        // Parsing failures or insufficient balance will trigger relevant errors.
+        // Empty input is allowed here (not a format error) — required validation is handled separately.
+        const isEmpty = fieldValue === '';
+        setErrors((prevErrors) => ({
+          ...prevErrors,
+          amount: '',
+        }));
+        if (!isEmpty) {
           try {
             const inputAmount = ethers.utils.parseUnits(
```
stanleyyconsensys, line 107: doesnt  the value may become NaN?

packages/wallet-ui/src/components/ui/organism/Header/SendInputModal/SendInputModal.view.tsx
stanleyyconsensys, line 119:
```suggestion
```

packages/wallet-ui/src/components/ui/organism/Header/SendInputModal/SendInputModal.view.tsx
```diff
@@ -94,7 +94,15 @@ export const SendInputModalView = ({
     }));
     switch (fieldName) {
       case 'amount':
-        if (fieldValue !== '' && fieldValue !== '.') {
+        // If input is not empty, attempt to parse and validate against balance.
+        // Parsing failures or insufficient balance will trigger relevant errors.
+        // Empty input is allowed here (not a format error) — required validation is handled separately.
+        const isEmpty = fieldValue === '';
+        setErrors((prevErrors) => ({
```
stanleyyconsensys, line 101: perhaps we should explain what is the purpose of this line (e.g reset error ...)
and what is the reason to reset error

stanleyyconsensys (changes requested):

packages/wallet-ui/src/components/ui/organism/Header/SendInputModal/SendInputModal.view.tsx
khanti42: parseUnits should always return a BigNumber otherwise throws

stanleyyconsensys (approved):
LGTM

PR #535, opened 2025-03-17
feat: unique account name in wallet ui

khanti42:
This PR is to ensure accountName uniqueness in Wallet-ui.
Uniqueness in wallet-ui is here: #532


stanleyyconsensys (approved):
LGTM

packages/wallet-ui/src/components/ui/organism/AccountDetailsModal/AccountDetailsModal.view.tsx
```diff
@@ -52,6 +60,18 @@ export const AccountDetailsModalView = () => {
       setIsEditing(false);
       return;
     }
+
+    // Check if the account name already exists
+    const accountExists = accounts.some(
+      (account) => account.accountName === trimedAccountName,
```
stanleyyconsensys, line 66: this is for updating , it seem we didnt exclude current accounts?

stanleyyconsensys (changes requested):

packages/wallet-ui/src/components/ui/organism/AccountDetailsModal/AccountDetailsModal.view.tsx
stanleyyconsensys: then may be adding back the trim

stanleyyconsensys (approved):
LGTM

PR #537, opened 2025-03-19
feat: search account by name in account list

khanti42:
This PR adds a search by name field in the account list.


packages/wallet-ui/src/components/ui/organism/AccountListModal/AccountList.view.tsx
```diff
@@ -46,16 +50,37 @@ export const AccountListModalView = ({
     await switchSnapAccount(chainId, account.address);
   };
 
+  // Filter visible accounts based on search term
+  const filteredVisibleAccounts = visibleAccounts.filter((account) =>
```
stanleyyconsensys, line 54: do it with useMemo, to avoid the filter again and again when state change becoz of other reason

stanleyyconsensys (changes requested):

packages/wallet-ui/src/components/ui/organism/AccountListModal/AccountList.view.tsx
khanti42: - Made the change for useMemo. 

- Regarding the handle for no account, filter is not called for empty array by design : https://www.w3schools.com/jsref/jsref_filter.asp 

- For the scroll bar, the effect is that the scroll is triggered only when opening the modal. Not when using the filtering. Which looks fine to me. wdyt ?

stanleyyconsensys: i mean UI handle for no accounts -> like display message

for scroll bar, if u verify the scroll bar only scroll bar when item is overflow , then it is fine, 
e.g if u have 100 account, after filiter , it becomes 99 accounts, u still need scroll bar

khanti42: I was meaning about the "auto scroll feature". The scroll bar stays if the number of element is bigger than the size of the box

khanti42: I updated the UX: the search box is hidden using CSS (with visibility: hidden) when there are no hidden accounts, but it’s not removed from the layout, so it still takes up space. This avoids a UI glitch that would occur when toggling between hidden and visible accounts.
- If hiddenAccounts.length === 0, we show: “No Hidden Accounts”.
- If hiddenAccounts.length > 0 but filteredHiddenAccounts.length === 0, we show: “No accounts found for the given search query”.

For visibleAccounts, we don’t handle this case because there’s always at least one visible account by definition, so we can always show "No accounts found for the given search query" if filteredVisibleAccounts.length === 0.

stanleyyconsensys (dismissed):
LGTM, 1 nit comment, optional

stanleyyconsensys (approved):
LGTM

PR #538, opened 2025-03-19
feat: account list contextual menu with hide/unhide and copy address feature

khanti42:
This PR enhances the Account List Modal by introducing a contextual menu that allows users to hide/unhide accounts and copy addresses.

https://github.com/user-attachments/assets/0744e41c-5bd0-4138-acd5-61e3fe1098ae

### DIFF
--- a/packages/wallet-ui/src/components/ui/organism/AccountListModal/AccountItem.view.tsx
+++ b/packages/wallet-ui/src/components/ui/organism/AccountListModal/AccountItem.view.tsx
@@ -1,5 +1,5 @@
-import { IconButton } from '@mui/material';
-
+import { IconButton, Menu } from '@mui/material';
+import { useState } from 'react';
 import { Account } from 'types';
 import { formatAddress } from 'utils/utils';
 import {
@@ -7,7 +7,9 @@ import {
   AccountImageStyled,
   VisibilityIcon,
   Wrapper,
+  MenuItem,
 } from './AccountItem.style';
+import { useMultiLanguage } from 'services';
 
 export interface Props {
   account: Account;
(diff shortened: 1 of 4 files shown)


packages/wallet-ui/src/components/ui/organism/AccountListModal/AccountItem.view.tsx
```diff
@@ -28,12 +30,12 @@ export const AccountItem = ({
   onItemClick,
   onIconButtonClick,
 }: Props) => {
+  const { translate } = useMultiLanguage();
   const { address, accountName } = account;
+  const [menuAnchorEl, setMenuAnchorEl] = useState<null | HTMLElement>(null);
 
   const preventDefaultMouseEvent = (event: React.MouseEvent) => {
-    // Prevent triggering the native behaviour
```
stanleyyconsensys, line 34: dont remove the comment if possible

packages/wallet-ui/src/components/ui/organism/AccountListModal/AccountItem.view.tsx
```diff
@@ -51,6 +54,20 @@ export const AccountItem = ({
     }
   };
 
+  const handleCopyAddress = () => {
+    navigator.clipboard.writeText(address);
+    setMenuAnchorEl(null); // Ferme le menu après la copie
```
stanleyyconsensys, line 59: please put comment in english

add toast to notify if the copy success? not sure if we have the same in account details

stanleyyconsensys (changes requested):
some nit comment

packages/wallet-ui/src/components/ui/organism/AccountListModal/AccountItem.view.tsx
khanti42: The point was that in the sidebar there is just a button so its not clear what it does, this is why showing "Copied!" when clicking is good idea. But in the dropdown in the list we explicitly say "Copy to clibpboard!" which makes the intent clear. And the copy cannot fail I guess, wdyt ?

stanleyyconsensys: agree to dismiss the fail case handle and notification for successful copy

stanleyyconsensys (approved):
LGTM

PR #539, opened 2025-03-21
fix: drop transactions from state if failed on blockchain

khanti42:
## Description

This PR updates the `getTransactions` method in the `TransactionService` and introduces logic to handle both failed and successful transactions in a single iteration. Failed transactions are now properly filtered out from the state, ensuring they do not persist unnecessarily. This improves efficiency and maintains the integrity of the transaction list.


stanleyyconsensys (changes requested):
This is requested by Starknet to include the tx that failed

what is the main reason to bypass those tx

khanti42:
Done elsewhere: https://github.com/Consensys/starknet-snap/pull/540

PR #540, opened 2025-03-26
fix: show transaction if failed

khanti42:
## Description

This PR solves an issue where reverted transaction were filtered out by the transaction service. 

## What changes with this PR

 - failed transaction now show up on the wallet-ui.
 - failed transaction that where staying as received (transactionOnState not updated) are now correctly updated with the status from blockchain


stanleyyconsensys (approved):
LGTM

PR #541, opened 2025-03-27
fix: account list mismatch when reconnecting Snap

khanti42:
## PR Description

The PR introduces a new Redux action: `setNetworksAndActiveNetwork({ networks, activeNetwork })`
- Replaces the two separate dispatches with a single atomic update to both networks and activeNetwork.
- Ensures consistent and up-to-date state when reconnecting to the Snap.
- Avoids race conditions or unnecessary re-renders that could happen between two separate dispatches.


stanleyyconsensys (approved):
LGTM, so do we still need "setNetworks" ?

stanleyyconsensys (approved):
LGTM

PR #543, opened 2025-04-01
fix: snap not working on firefox

khanti42:
This PR fixes Snap installation and usage on Firefox latest MetaMask. 

## Issue 1: In firefox calling `invokeSnap` with undefined fails, so we change this 

```
          request: {
            method,
            params: params ? removeUndefined(params) : params,
          },
```

by this 

```
          request: !params
            ? { method }
            : { method, params: params ? removeUndefined(params) : params },
``` 

## Issue 2: In firefox calling invokeSnap with an [Object] fails. We need to stringily and parse it with json to get a real plain object. 

We add `JSON.parse(JSON.stringify( .... ))` before returning in the removeUndefined utility function.

### DIFF
--- a/packages/wallet-ui/src/services/useSnap.ts
+++ b/packages/wallet-ui/src/services/useSnap.ts
@@ -40,10 +40,9 @@ export const useSnap = () => {
         method: 'wallet_invokeSnap',
         params: {
           snapId,
-          request: {
-            method,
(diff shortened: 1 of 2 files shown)


stanleyyconsensys (approved):
LGTM, nice catch

PR #544, opened 2025-04-01
chore: fix translation wrong keys

khanti42:
There is a typo we use `translate('signeMessagePrompt')` but it should be `translate('signMessagePrompt')`

signeMessagePrompt => signMessagePrompt


stanleyyconsensys (approved):
LGTM

PR #547, opened 2025-04-06
chore: fix max fees and usd toggle

khanti42:
This PR is to fix two things: 

- The USD toggle button was not showing the right amount in USD
- The max amount to be spend was not accounting for the feeToken in the calculation causing max amount error on validation screen


stanleyyconsensys (commented):
LGTM, just 1 minor comment

PR #556, opened 2025-07-14
fix: incorrect truncation in getMaxDecimalsReadable

khanti42:
## 🐛 Fix incorrect truncation in `getMaxDecimalsReadable`

This PR fixes a bug where integer values like `"100"` were incorrectly truncated to `"1"` due to logic that assumed all inputs had a decimal point. The function now properly checks for the presence of a decimal and avoids applying decimal trimming logic to whole numbers. This ensures values like `"100"` or `"10"` are returned as-is, while still correctly trimming and formatting decimal values.


mario-christopher:
Can we add a Unit Test to test this with suitable edge conditions ?

mario-christopher (approved):

PR #581, opened 2026-08-18
fix: starkscan api integration, update response structure and fix timestamp handling 

khanti42:
> [!NOTE]
> **Medium Risk**
> Changes affect how wallet transaction history is fetched and normalized; incomplete v1 mapping (synthetic invokes, missing fees/revert detail) could skew displayed activity until further API fields are wired.
> 
> **Overview**
> **StarkScan transaction fetching** is updated from the legacy `/api/v0` contract endpoints to **`https://api.starkscan.co/v1/{SN_SEPOLIA|SN_MAIN}/address/{address}/transactions`**, with **`nextCursor`**-based pagination instead of **`next_url`**, and auth via **`X-Starkscan-Api-Key`** (Sepolia no longer uses a separate host).
> 
> Response types and mapping are aligned with the new payload: list fields **`items`** / **`nextCursor`**, per-tx **`txHash`**, **`txType`**, **`timestampIso`**, **`fromAddress`** / **`toAddress`**, and transfer summary fields. **`toTransaction`** now derives Unix time from ISO strings, treats deploy vs invoke from **`txType`**, and builds invoke **`accountCalls`** from **`topTransferTokenAddress`**, **`counterparty`**, **`topTransferAmount`**, and **`primaryMethod`** rather than **`account_calls`**. Fees and version are largely **stubbed** (`maxFee`/`actualFee`, fixed version **3**) where the v1 shape no longer exposes the old fields.
> 
> Shared **`ApiClient`** only sets **`Content-Type: application/json`** on **POST** requests so GET StarkScan calls are not sent with a JSON content type. Tests and **`generateStarkScanTransactions`** mocks were rewritten to match the v1 schema (fixture JSON import removed).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1f86a0be432e51bacf54fb41fd7b2ff0e3f8bc87. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/starknet-snap/src/chain/data-client/starkscan.ts
+++ b/packages/starknet-snap/src/chain/data-client/starkscan.ts
@@ -207,7 +207,14 @@ export class StarkScanClient extends ApiClient implements IDataClient {
 
     let transaction: V2Transaction;
 
-    if (!this.isDeployTransaction(tx)) {
+    if (this.isDeployTransaction(tx)) {
+      transaction = newDeployTransaction({
+        txnHash,
+        senderAddress,
+        chainId,
+        txnVersion,
+      });
+    } else {
       const tokenAddress = topTransferTokenAddress ?? tx.toAddress ?? '';
       const calldata =
         counterparty && topTransferAmount
@@ -227,13 +234,6 @@ export class StarkScanClient extends ApiClient implements IDataClient {
               },
             ]
           : [],
-        txnVersion,
-      });
-    } else {
-      transaction = newDeployTransaction({
-        txnHash,
-        senderAddress,
-        chainId,
         txnVersion,
       });
     }
--- a/packages/starknet-snap/src/chain/data-client/starkscan.ts
+++ b/packages/starknet-snap/src/chain/data-client/starkscan.ts
@@ -27,32 +27,44 @@ export class StarkScanClient extends ApiClient implements IDataClient {
 
   protected options: StarkScanOptions;
 
-  protected deploySelectorName = 'constructor';
-
   constructor(network: Network, options: StarkScanOptions) {
     super();
     this.network = network;
     this.options = options;
   }
 
-  protected get baseUrl(): string {
+  protected get chain(): string {
     switch (this.network.chainId) {
       case constants.StarknetChainId.SN_SEPOLIA:
-        return 'https://api-sepolia.starkscan.co/api/v0';
+        return 'SN_SEPOLIA';
       case constants.StarknetChainId.SN_MAIN:
-        return 'https://api.starkscan.co/api/v0';
+        return 'SN_MAIN';
       default:
         throw new InvalidNetworkError();
     }
   }
 
-  protected getApiUrl(endpoint: string): string {
-    return `${this.baseUrl}${endpoint}`;
+  protected get baseUrl(): string {
+    switch (this.network.chainId) {
+      case constants.StarknetChainId.SN_SEPOLIA:
+      case constants.StarknetChainId.SN_MAIN:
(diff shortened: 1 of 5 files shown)


packages/starknet-snap/src/chain/data-client/starkscan.ts
```diff
@@ -27,32 +27,44 @@ export class StarkScanClient extends ApiClient implements IDataClient {
...
+    }
+  }
+
   protected get baseUrl(): string {
     switch (this.network.chainId) {
       case constants.StarknetChainId.SN_SEPOLIA:
-        return 'https://api-sepolia.starkscan.co/api/v0';
       case constants.StarknetChainId.SN_MAIN:
-        return 'https://api.starkscan.co/api/v0';
+        return 'https://api.starkscan.co';
```
stanleyyconsensys, line 51: all using `https://api.starkscan.co` regardless network?

packages/starknet-snap/src/chain/data-client/starkscan.ts
```diff
@@ -78,26 +90,20 @@ export class StarkScanClient extends ApiClient implements IDataClient {
...
    * @param address - The address of the contract to fetch the transactions for.
-   * @param to - The filter includes transactions with a timestamp that is >= a specified value, but the deploy transaction is always included regardless of its timestamp.
+   * @param to - Unix timestamp; include txs at or after this time. Deploy txs are always included.
    * @returns A Promise that resolve an array of Transaction object.
    */
   async getTransactions(address: string, to: number): Promise<Transaction[]> {
-    let apiUrl = this.getApiUrl(
-      `/transactions?contract_address=${address}&order_by=desc&limit=${this.limit}`,
-    );
+    let apiUrl = this.getApiUrl(address);
```
stanleyyconsensys, line 100: just question, we always use the same path for all use case, is that correct

stanleyyconsensys (dismissed):
not a block, just question
i give approve first, but if the comment is valid, plz drop the approve

packages/starknet-snap/src/chain/data-client/starkscan.ts
khanti42: Yes. The new Starkscan public API is one host: `https://api.starkscan.co`. Network is no longer a subdomain (`api-sepolia` vs `api`); it is the `{chain}` path segment.

- Sepolia: `https://api.starkscan.co/v1/SN_SEPOLIA/address/{address}/transactions`
- Mainnet: `https://api.starkscan.co/v1/SN_MAIN/address/{address}/transactions`

That matches [Starkscan’s API docs](https://starkscan.co/docs/api) (`Production base = https://api.starkscan.co`, resources `/v1/{chain}/...`). `this.chain` is `SN_SEPOLIA` or `SN_MAIN`; anything else still throws `InvalidNetworkError`.

packages/starknet-snap/src/chain/data-client/starkscan.ts
khanti42: Yes, exactly. v1 only exposes one address-tx list: `GET /v1/{chain}/address/{address}/transactions`. There is no separate deploy-tx route like the old v0 query.

So both `getTransactions` and `getDeployTransaction` call `getApiUrl(address)`:
- `getTransactions` paginates that list (`cursor`) until the time window is covered, and still includes deploy txs.
- `getDeployTransaction` hits the same first page and picks the `DEPLOY_ACCOUNT` item if the list did not already include it.

stanleyyconsensys (approved):

PR #582, opened 2026-08-18
feat: deprecation notice for new account

khanti42:
https://github.com/user-attachments/assets/061e29a9-6146-495b-a1af-c2596f293a2f

> [!NOTE]
> **Medium Risk**
> Blocks account deployment and changes execute-txn behavior for undeployed users, but deployed-account paths are unchanged and the change is intentional product deprecation.
> 
> **Overview**
> **New Starknet accounts are no longer supported.** Undeployed or not-yet-confirmed-on-chain accounts get a deprecation message and a link to [starknet.io/wallets](https://www.starknet.io/wallets) instead of the normal wallet experience.
> 
> In the **snap**, the home page only shows a copyable address, balance, and network when deployment is verified on-chain (`shouldShowAccountAddress` fails closed on errors). Transaction confirmation no longer runs deploy-included flows: if fees imply account deployment (`includeDeploy`), it shows `showAccountCreationDeprecatedModal` and rejects with `UserRejectedOpError` instead of opening the execute UI or calling `deployAccount`.
> 
> In **wallet-ui**, when `isDeployed !== true`, the home page shows a warning banner and hides header/transactions; the sidebar drops copy/explorer/account-details affordances; account list and drawer hide formatted addresses and “copy address” for undeployed entries.
> 
> EN/FR strings were added in both packages. Wallet `constants` also adds `STARKNET_WALLETS_URL` and extra USDC token entries for Coingecko on mainnet/Sepolia.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ee991577d6252233d484e8db582d829dfbc6fbb4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/starknet-snap/src/rpcs/execute-txn.ts
+++ b/packages/starknet-snap/src/rpcs/execute-txn.ts
@@ -24,6 +24,7 @@ import {
 } from '../utils';
 import { CAIRO_VERSION } from '../utils/constants';
 import { UserRejectedOpError } from '../utils/exceptions';
+import { showAccountCreationDeprecatedModal } from '../utils/snapUtils';
 import {
   deployAccount,
   executeTxn as executeTxnUtil,
@@ -247,6 +248,11 @@ export class ExecuteTxnRpc extends AccountRpcController<
     resourceBounds,
     includeDeploy,
   }: ConfirmTransactionParams): Promise<TransactionRequest> {
+    if (includeDeploy) {
+      await showAccountCreationDeprecatedModal();
+      throw new UserRejectedOpError() as unknown as Error;
+    }
+
     const requestId = uuidv4();
     const { chainId, name: networkName } = this.network;
     const { hdIndex: addressIndex } = this.account;
--- a/packages/starknet-snap/locales/en.json
+++ b/packages/starknet-snap/locales/en.json
@@ -1,6 +1,12 @@
 {
   "locale": "en",
   "messages": {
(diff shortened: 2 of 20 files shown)


packages/starknet-snap/src/rpcs/execute-txn.ts
```diff
@@ -247,6 +248,11 @@ export class ExecuteTxnRpc extends AccountRpcController<
     resourceBounds,
     includeDeploy,
   }: ConfirmTransactionParams): Promise<TransactionRequest> {
+    if (includeDeploy) {
+      await showAccountCreationDeprecatedModal();
+      throw new UserRejectedOpError() as unknown as Error;
```
stanleyyconsensys, line 253: just question

any reason we throw user reject?

stanleyyconsensys (approved):

packages/starknet-snap/src/rpcs/execute-txn.ts
khanti42: No specific reason — just to avoid introducing a new error type and keep things simple. Mostly to abort cleanly after the alert. UserRejectedOpError is our existing cancel path and maps to UserDeny, so clients (e.g. wallet-ui’s isUserDenyError) swallow it instead of showing a second error.

Montoya:
lgtm

